PR #42, opened 2024-02-21
Add changeSelectedTab

mk-fan:
## Problem
- Need a way to change the currently selected tab via an API

## Solution
- Add changeSelectedTab

## License


dogusata (commented):
This change will not be implemented. You cannot change tabs on behalf of user.

Please describe the reason why it is necessary, UX flow decisions and the use case please.

Expected steps are:
- Create a feature request, describe the reason
- Wait till the final decision happen on the discussion
- Depending on the decision, you can contribute or wait for the implementation.

mk-fan:
After discussing it, the feature was included in https://github.com/aws/mynah-ui/commit/0844cc1943ce0bfcc89ef7ce3845f72689bb6d80.

PR #47, opened 2024-05-02
Dogusata/added development guidelines and api docs

dogusata:
## Problem
MynahUI lacks a proper Developer guideline to let people contribute to the project. Also there is no api doc available.

## Solution
- Added a guide to add a component from scratch to `mynah-ui`
- Added API docs generator and deployment procedure

## License


docs/DEVELOPER.md
```diff
@@ -0,0 +1,499 @@
+## Hey developer!
+
+Would you like to improve, or add new components to `mynah-ui`? Then here's a guide for you, which will both tell the basics, rules and a complete guide to a functional component additions.
+
+### Before you start!
+
+__There are a few key things you need to keep in mind all the time.__
+
+- As you can see from the project type on github _(which is a guess by github)_, it might look like `mynah-ui` is an `SCSS` project. And to be fair, it is not completely wrong. We're trying to solve as much as we can with css. On the typescript/code side, we're just generating dom elements, making calculations and more importantly: toggling css classes on and off for the elements we create.
```
justinmk3, line 9: > We're trying to solve as much as we can with css

Very useful insight. Helps other devs understand the strategy/positioning of this library.

CONTRIBUTING.md
```diff
@@ -21,6 +21,8 @@ reported the issue. Please try to include as much information as you can. Detail
 
 
 ## Contributing via Pull Requests
+**Before start; please do follow the [DEVELOPER Guidlines](./docs/DEVELOPER.md) first.**
```
justinmk3, line 24: could be much more prominent
```suggestion

**Read [DEVELOPER Guidelines](./docs/DEVELOPER.md) first.**
```

I would even repeat this at the top of CONTRIBUTING.md. It's quite important and much more useful that the boilerplate that is currently in CONTRIBUTING.md. (p.s.: can CONTRIBUTING.md be reduced? is all of its text essential and useful for most devs? if not, they may miss the actually important parts)

justinmk3 (approved):
Great insights. 

I would consider dropping `typedoc`, unless you think those docs will be useful for search engines or referencing in issues/discussions.

docs/DEVELOPER.md
dogusata: Thanks🙏
In later, we're also trying to inform the developers that if we see any kind of inline styles etc, it will be rejected ;)

volodkevych (approved):

dogusata:
> Great insights.
> 
> I would consider dropping `typedoc`, unless you think those docs will be useful for search engines or referencing in issues/discussions.

For this, with every release, we'll put an updated version of the api docs [under a folder on the live demo github page](https://aws.github.io/mynah-ui/api-doc/index.html). Which i believe will be very helpful to the `mynah-ui` consumers to follow up with the data model changes/additions or provided method additions etc.

PR #50, opened 2024-05-24
Dogusata/fix update chat answer with message id not working

dogusata:
## Problem
To update a chat card, if it is/was the latest streaming card, it can be updated with `updateLastChatItem` method. For all other cases `updateChatItemWithMessageId` should work. However, if there is no streaming card at all in the whole window, cards are not getting updated with both of the functions.

#### Expected Behaviour:
Doesn't matter what type is the chat item and it is the last of not, it has to be updated with `updateChatItemWithMessageId` if the chat item with the given messageId is rendered already.

#### Steps to Reproduce:
Create a card by adding a new ChatItem with type `ANSWER` without providing the `body` attribute. After that try to update it with any of the mentioned methods above.

## Solution
The checks for adding a card render to the rendered items list was insufficient. Any card (if they are rendered) should be added to the rendered items list to allow them to be updated. Previously the UX decision was to not update the cards after they finish the streaming. However for cases like WeaverBird's file rejection and reverting rejects, previous cards should also allow updating.

## License

### DIFF
--- a/example/src/main.ts
+++ b/example/src/main.ts
@@ -192,19 +192,19 @@ export const createMynahUI = (initialData?: MynahUIDataModel): MynahUI => {
       Log(`File clicked: <b>${filePath}</b>`);
     },
     onFileActionClick: (tabId, messageId, filePath, actionName) => {
+      console.log("YARAK!");
       Log(`File action clicked: <b>${filePath}</b> -> ${actionName}`);
       switch (actionName) {
-        case 'update-comment':
-        case 'comment-to-change':
-          showCustomForm(tabId);
-          break;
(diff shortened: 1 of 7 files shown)


example/src/main.ts
```diff
@@ -192,19 +192,18 @@ export const createMynahUI = (initialData?: MynahUIDataModel): MynahUI => {
       Log(`File clicked: <b>${filePath}</b>`);
     },
     onFileActionClick: (tabId, messageId, filePath, actionName) => {
+      console.log("YARAK!");
```
rtarcr, line 195: does this need to be here?

PR #55, opened 2024-06-13
Dogusata/add code diff to syntax highlighter and file description tooltip

dogusata:
## Problem
- SyntaxHighlighter should support diff view. Since it is already there with `prism.js` it has to be enabled.
- FileItem component should support markdown tooltips
- Sometimes tooltips exceeding the screen edges
## Solution
- Updated SyntaxHighlighter component to support diff view too
- Added tooltip ability to FileItem component
- Fixed tooltip edge bleeding issue.

## License

### DIFF
--- a/src/components/chat-item/chat-item-tree-file.ts
+++ b/src/components/chat-item/chat-item-tree-file.ts
@@ -1,8 +1,12 @@
+import { marked } from 'marked';
 import { DomBuilder, ExtendedHTMLElement } from '../../helper/dom';
 import { MynahUIGlobalEvents, cancelEvent } from '../../helper/events';
 import { FileNodeAction, MynahEventNames, TreeNodeDetails } from '../../static';
 import { Button } from '../button';
+import { Card } from '../card/card';
+import { CardBody } from '../card/card-body';
 import { Icon, MynahIcons } from '../icon';
+import { Overlay, OverlayHorizontalDirection, OverlayVerticalDirection } from '../overlay';
 
 export interface ChatItemTreeFileProps {
   tabId: string;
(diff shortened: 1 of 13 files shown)


src/components/chat-item/chat-item-tree-file.ts
```diff
@@ -96,4 +114,39 @@ export class ChatItemTreeFile {
       ]
     });
   }
+
+  private readonly showTooltip = (content: string, vDir?: OverlayVerticalDirection, hDir?: OverlayHorizontalDirection): void => {
+    if (content.trim() !== undefined) {
```
kmile, line 119: When will `String.trim` ever return `undefined`?

src/components/chat-item/chat-item-tree-file.ts
```diff
@@ -96,4 +114,39 @@ export class ChatItemTreeFile {
...
+            }).render
+          ],
+        });
+      }, PREVIEW_DELAY);
+    }
+  };
+
+  public readonly hideTooltip = (): void => {
+    clearTimeout(this.fileTooltipTimeout);
+    if (this.fileTooltip !== null) {
```
kmile, line 147: Is this `if` necessary? The conditional code paths only differ between `null` and `undefined` now, but the effect of both is effectively the same.

dogusata: We're manually setting it to null, but would be better to make it as `this.fileTooltip != null`. Thanks for catching it up.

src/components/chat-item/chat-item-tree-file.ts
```diff
@@ -143,8 +143,10 @@ export class ChatItemTreeFile {
   };
 
   public readonly hideTooltip = (): void => {
-    clearTimeout(this.fileTooltipTimeout);
-    if (this.fileTooltip !== null) {
+    if (this.fileTooltipTimeout != null) {
+      clearTimeout(this.fileTooltipTimeout);
+    }
+    if (this.fileTooltip != null) {
```
kmile, line 149: Not blocking, but I still don't think these `if` statements have any effect?

kmile (dismissed):

src/components/chat-item/chat-item-tree-file.ts
dogusata: Hmm, hideTooltip can be called because of the mouse leave/out event fired while user passing their mouse over the screen. Since because of the tooltip delay, when the hideTooltip fired the tooltip component might not be there yet, which may cause an error to throw.

dogusata: Done

PR #57, opened 2024-06-18
feature: scss mixins and efficiencies

32teeth:
## Problem
SCSS was very repetitive and could use some improvement 

## Solution

- [x] introduce new `@mixin`s
- [x] use said mixins in `_variables`
- [x] introduce `@for` and `@each` native SCSS iterator strategies
- [x] introduce 1D and 2D arrays in SCSS

the above provides for scalability, repurposability and predictability in SCSS

### To Do

Once merged we can explore more SCSS efficiencies and structure this into being a design system based record or work

## License

### DIFF
--- a/src/styles/_variables.scss
+++ b/src/styles/_variables.scss
@@ -73,9 +73,9 @@
      * Status variables
      */
     $mynah-status-colors: (
-        'info': #0971d3
-        'success': #037f03
-        'warning': #b2911c
+        'info': #0971d3,
+        'success': #037f03,
+        'warning': #b2911c,
         'error': #d91515
     );
 
--- a/src/styles/_variables.scss
+++ b/src/styles/_variables.scss
@@ -6,34 +6,33 @@
     --mynah-max-width: 2560px;
     --mynah-sizing-base: 0.25rem;
     --mynah-sizing-half: calc(var(--mynah-sizing-base) / 2);
-    --mynah-sizing-1: var(--mynah-sizing-base);
-    --mynah-sizing-2: calc(var(--mynah-sizing-base) * 2);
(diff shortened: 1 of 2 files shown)


justinmk3:
Do we actually need SCSS or could we use plain CSS? That would avoid an extra layer / build step / dependency. And modern CSS reduces the need for SCSS, these days.

32teeth:
> Do we actually need SCSS or could we use plain CSS? That would avoid an extra layer / build step / dependency. And modern CSS reduces the need for SCSS, these days.

Hi @justinmk3 

Yes, modern CSS, including new (recent) module nesting and rule setting. Although an interesting insight is CSS nesting is different from CSS preprocessors such as [Sass](https://sass-lang.com/) in that it is parsed by the browser rather than being pre-compiled by a CSS preprocessor. 

> [!NOTE]
[mynah-ui](https://github.com/aws/mynah-ui) is already uses SCSS preprocessors. 

> [!TIP]
This PR, enhances and extends the use of variable declaration using `@each` and `@for`

src/styles/_variables.scss
```diff
@@ -6,34 +6,33 @@
...
-    --mynah-sizing-14: calc(var(--mynah-sizing-base) * 14);
-    --mynah-sizing-15: calc(var(--mynah-sizing-base) * 15);
-    --mynah-sizing-16: calc(var(--mynah-sizing-base) * 16);
-    --mynah-sizing-17: calc(var(--mynah-sizing-base) * 17);
-    --mynah-sizing-18: calc(var(--mynah-sizing-base) * 18);
+
+    /**
+     * Sizing variables
+     */
+    @for $i from 1 through 18 {
```
dogusata, line 13: Lovely!!

PR #62, opened 2024-06-20
(feat): Add context selector to prompt input

dogusata:
## Problem
Exactly like the quick action commands selector, a context selector should be implemented into the prompt input field. But unlike the quick action commands, it can be picked multiple times and should be written inside the prompt text. It should also send the selected context items in an additional attribute in the `prompt` object. 
The available context items should be provided through the data model like the QuickAction commands but in a new attribute field.

We want to allow users to specify a target for the context like `workspace` or in the future `file` or `folder`. 

## Solution
- A new field added in the tabStore data model which is called `contextCommands` and accepts an array of `QuickActionCommandGroup`
- Prompt input text field listens `@` character insertion and enables the overlay picker for the provided context items if there is any.
- Prompt input checks if the letter trying to be removed with DEL or BACKSPACE, it checks if the cursor position is in a word which is a mention (starts with `@`) and removes the whole word/mention.
- All DOCS are updated with the according information of `contextCommands` addition.
- Example app is also updated to showcase the context selector functionality.

## License


src/components/chat-item/prompt-input/prompt-text-input.ts
```diff
@@ -112,13 +112,53 @@ export class PromptTextInput {
...
+    let prevSpaceIndex = -1;
+    let nextSpaceIndex = currentValue.indexOf(' ', cursorPos);
+
+    // We're not splitting the text value by spaces to get the words and check all of them
+    // Reason behind that is performance concerns.
+    // We know that we're looking for a word, and we only need the word for the given index if it is inside a word
+
+    // Find previous space chararacter
+    for (let i = cursorPos - 1; i >= 0; i--) {
+      if (currentValue[i] === ' ') {
```
imykhai, line 147: probably nitpicking, but this will miss a word that is separated by other [whitespace characters](https://en.wikipedia.org/wiki/Whitespace_character) or quotes etc.

imykhai (approved):

src/components/chat-item/prompt-input/prompt-text-input.ts
dogusata: Should not effect finding the word start index, since we're just trying to find the previous space if there is any.

ege0zcan (commented):
I wasn't able to finish my review before this got merged, I only noticed small typos in the PR

PR #64, opened 2024-06-21
Dogusata/fix stream animation flows and card orders

dogusata:
## Problem
- `<span class='typewriter-part>...</span>` shows up in contents within list items.
- Chat card component reordering and update structure is broken sometimes depending on update type.

## Solution
- Streaming content component wrapped out from the chat-item-card and builded as a separate component
- Typewriter animation injection removed from the markdown parsing process
- Typewriter animation injection added to node process state with a check if the node type is TEXT, then we're adding the animation wrapper.
- Converted some components to exportable ones.
- Moved related `scss` file loads to their own or related components
- Updated some component's code structures to match them repo wide

## License


src/styles/components/chat/_chat-prompt-attachment.scss
```diff
@@ -1,3 +1,4 @@
+@import '../../mixins';
```
ege0zcan, line 1: nit: is this import used in this css file

src/styles/components/chat/_chat-item-tree-view.scss
```diff
@@ -196,7 +196,7 @@
         flex-flow: row-reverse nowrap;
         justify-content: flex-end;
         align-items: center;
-        gap: var(--mynah-sizing-2);
+        // gap: var(--mynah-sizing-2);
```
ege0zcan, line 199: is this intentional or a leftover?

src/helper/events.ts
```diff
@@ -7,9 +7,9 @@ import { MynahEventNames } from '../static';
 import { generateUID } from './guid';
 
 export const cancelEvent = (event: Event): boolean => {
-  event.preventDefault();
-  event.stopPropagation();
-  event.stopImmediatePropagation();
+  event.preventDefault?.();
```
ege0zcan, line 10: Doesn't all events have these functions?

src/components/card/card-body.ts
```diff
@@ -97,80 +87,93 @@ export class CardBody {
...
-            click: (e: MouseEvent) => {
-              if (this.props.onLinkClick !== undefined) {
-                this.props.onLinkClick(url, e);
-              }
-            },
-            auxclick: (e: MouseEvent) => {
-              if (this.props.onLinkClick !== undefined) {
-                this.props.onLinkClick(url, e);
-              }
+    let elementFromNode: HTMLElement = node;
```
ege0zcan, line 90: nit:  this function is too big, let's consider putting logic into functions like:
```
processNode(){
 if(text.node)
   animateTextNode()
  else if(linkNode)
    linkNode()

}
```

src/components/syntax-highlighter.ts
```diff
@@ -259,8 +244,27 @@ export class SyntaxHighlighter {
...
+              ? [ {
+                  type: 'span',
+                  classNames: [ 'mynah-syntax-highlighter-language' ],
+                  children: [ props.language ]
+                } ]
+              : []),
+            ...this.codeBlockButtons
+          ],
+        }));
+      }, 1);
```
ege0zcan, line 266: nit: we should explain why we have this 1ms timeout

src/helper/events.ts
dogusata: Unfortunately not. It can be overridden during generation

src/components/card/card-body.ts
dogusata: Next plan is to move it to a helper function

ege0zcan (approved):
Went through changes together in a call, made small updates together. Looks good to me

PR #74, opened 2024-07-10
fix: restore command prompt when blurred then focused

32teeth:
## Problem

When a user types a quick action command starting with `/` or a context command starting with `@` and then loses focus, the command context is lost. When the user refocuses on the input field and continues typing, the quick pick overlay does not reappear, causing the command to never be picked up.

## Solution

- [x] Added a focus event listener to the PromptTextInput class to check the input value when the input field regains focus.
- [x] If the input value starts with / or @, the quick pick overlay is reopened, and the commands are filtered accordingly.
- [x] Ensured that the quick pick overlay behaves as expected when the user resumes typing after losing focus.
- [x] If a user previous typed half a command `/show` which filtered out 4 commands when they re-focus on the input field, only restore those 4 commands

## Preview

restoreCommandsOnFocus

## To Do

> [!TIP]
> Tomorrow Day Problems
* `chat-prompt-input.ts` could probably use some redundancy grouping for similar functionality
* `handleInputKeydown` has a complexity score of 54  

## License

### DIFF
--- a/src/components/chat-item/chat-prompt-input.ts
+++ b/src/components/chat-item/chat-prompt-input.ts
@@ -53,6 +53,7 @@ export class ChatPromptInput {
       tabId: this.props.tabId,
       onKeydown: this.handleInputKeydown,
       onInput: () => this.updateAvailableCharactersIndicator(),
+      onFocus: this.handleInputFocus,
     });
     this.remainingCharsIndicator = DomBuilder.getInstance().build({
       type: 'span',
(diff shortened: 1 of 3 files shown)


src/components/chat-item/prompt-input/prompt-text-input.ts
```diff
@@ -72,6 +73,9 @@ export class PromptTextInput {
         },
         focus: () => {
           this.render.addClass('input-has-focus');
+          if (this.props.onFocus !== undefined) {
```
meysius, line 76: if (this.props.onFocus !== undefined): This statement checks if this.props.onFocus is not equal to undefined. It will evaluate to true if this.props.onFocus is defined and has any value other than undefined. This includes values like null, 0, false, NaN, and an empty string ''. It will evaluate to false only if this.props.onFocus is strictly equal to undefined.

if (typeof this.props.onFocus !== 'undefined'): This statement checks if the type of this.props.onFocus is not equal to the string 'undefined'. It will evaluate to true if this.props.onFocus is defined and has any value, including null, 0, false, NaN, and an empty string ''. It will evaluate to false only if this.props.onFocus is strictly equal to undefined.

In practice, both statements will work the same way in most cases, as they both check if this.props.onFocus is defined or not.

However, there is a subtle difference when dealing with null. If this.props.onFocus is null, then this.props.onFocus !== undefined will evaluate to true, while typeof this.props.onFocus !== 'undefined' will also evaluate to true.

In general, it is considered better practice to use the typeof approach (typeof this.props.onFocus !== 'undefined') because it explicitly checks for the type of the variable, which can help prevent some edge cases and potential bugs.

src/components/chat-item/chat-prompt-input.ts
```diff
@@ -310,6 +311,55 @@ export class ChatPromptInput {
     }
   };
 
+  private readonly handleInputFocus = (): void => {
+    const inputValue = this.promptTextInput.getTextInputValue();
+    if (inputValue.startsWith('/') || inputValue.startsWith('@')) {
```
meysius, line 316: is there an @ trigger ?

src/components/chat-item/chat-prompt-input.ts
```diff
@@ -310,6 +311,55 @@ export class ChatPromptInput {
...
+          const searchTerm = inputValue.substring(this.quickPickTriggerIndex).match(/\S*/gi)?.[0];
+          const promptRegex = new RegExp(searchTerm ?? '', 'gi');
+          newQuickPickCommandGroup.commands = newQuickPickCommandGroup.commands.filter(command =>
+            command.command.match(promptRegex)
+          );
+          if (newQuickPickCommandGroup.commands.length > 0) {
+            restorePreviousFilteredQuickPickItemGroups.push(newQuickPickCommandGroup);
+          }
+        } catch (e) {
+          // In case the prompt is an incomplete regex
```
meysius, line 339: is there a logging mechanism to not swallow this silently?

meysius (commented):
Added some comments

src/components/chat-item/prompt-input/prompt-text-input.ts
dogusata: From typescript defs, it cannot get `any` kind of value.

src/components/chat-item/chat-prompt-input.ts
dogusata: Yes

dogusata:
There are **two issues** related with the `@` command picker in the PR:

### 1st issue
When user starts with @ to pick a context (currently this is the general expected behavior btw.) and blurs the input area and comes back in. There is an additional @ char.

Gif: (loading)
context-selector-problem-1

### 2nd issue
When there is another context picker trigger (second `@`) it is becomes not possible to write text again.

Gif: (loading)
context-selector-problem-2

**Note about how they are and should work:** 
Quick action commands can be opened only if there is no text written. Even after you type sth. you cannot go back to the beginning of the line and open it. Reason behind that is we cannot be sure that if the command accepts additional text or not. And there can be only one command selected. 

However, context selection (`@`) is different, it is just a predefined text selection which always starts with `@`. User can select them anywhere inside the text and pick as much as they want.

src/components/chat-item/prompt-input/prompt-text-input.ts
32teeth: Hmm, i only repurposed the one that existed for `input`

```
input: (e: KeyboardEvent) => {
  // Set the appropriate prompt input height
  this.updatePromptTextInputSizer();
  // Additional handler
  if (this.props.onInput !== undefined) {
    this.props.onInput(e);
  }
},
```

we could do

```
focus: () => {
  this.render.addClass('input-has-focus');
  if (typeof this.props.onFocus !== 'undefined') {
    this.props.onFocus();
  }
},
```

if that works, please upvote

32teeth:
> However, context selection (`@`) is different, it is just a predefined text selection which always starts with `@`. User can select them anywhere inside the text and pick as much as they want.

Ah, GTK! TY

src/components/chat-item/prompt-input/prompt-text-input.ts
```diff
@@ -72,6 +73,9 @@ export class PromptTextInput {
         },
         focus: () => {
           this.render.addClass('input-has-focus');
+          if (typeof this.props.onFocus !== 'undefined') {
```
dogusata, line 76: Even though type val check is sufficient, `this.props.onFocus != null` would be better i suppose.

dogusata: Not mandatory btw, just a nit for later.

dogusata:
> > However, context selection (`@`) is different, it is just a predefined text selection which always starts with `@`. User can select them anywhere inside the text and pick as much as they want.
> 
> Ah, GTK! TY

Checks, works as expected 👍🏻

dogusata (approved):
Works as expected.

src/components/chat-item/chat-prompt-input.ts
dogusata: Not yet, will be there in the future if it is being enabled from the configs.

src/components/chat-item/prompt-input/prompt-text-input.ts
dogusata: This will work also just `!= null` check should be fine, but not mandatory or urgent.

32teeth:
Any hesitation on merge?

PR #80, opened 2024-07-18
fix(tabs): Make max tabs notification dismissible and allow external configuration

tverney:
## Problem

Currently the notification for max tabs isn't dismissible. Also, isn't possible the implementation overwrite the defaults (duration / tooltip text).

notification

## Solution

Add maxReachedOverlayDuration (to adjust the duration / fallbacks to default 5000) to use as a dismiss duration and noMoreTabsTooltip exposable for messaging configuration.

notification

Note: I've tried to update the changelog with newChange, but doesn't seems to have the script.

## License

### DIFF
--- a/src/components/navigation-tabs.ts
+++ b/src/components/navigation-tabs.ts
@@ -16,10 +16,13 @@ import { TabBarButtonsWrapper } from './navigation-tab-bar-buttons';
 import { Overlay, OverlayHorizontalDirection, OverlayVerticalDirection } from './overlay';
 import { Toggle, ToggleOption } from './toggle';
 import '../styles/components/_nav-tabs.scss';
+import { DEFAULT_TIMEOUT } from './notification';
 
 export interface TabsProps {
   onChange?: (selectedTabId: string) => void;
+  noMoreTabsTooltip?: string;
   onBeforeTabRemove?: (tabId: string) => boolean;
+  maxReachedOverlayDuration?: number;
 }
 export class Tabs {
   render: ExtendedHTMLElement;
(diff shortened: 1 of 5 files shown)


src/components/navigation-tabs.ts
```diff
@@ -155,6 +158,16 @@ export class Tabs {
         }).render
       ],
     });
+
+    if (duration !== undefined && duration !== -1) {
```
ege0zcan, line 162: in which case can this be `-1`

src/static.ts
```diff
@@ -431,6 +431,8 @@ export interface ConfigOptions {
   }>;
   tabBarButtons?: TabBarMainAction[];
   maxTabs: number;
+  maxReachedOverlayDuration?: number;
```
ege0zcan, line 434: what do you think about naming this as `maxTabsTooltipDuration`  I feel like the variable should contain `tabs` in the name to make it easier to reason with

src/components/notification.ts
```diff
@@ -28,7 +30,7 @@ export class Notification {
   private readonly props;
 
   constructor (props: NotificationProps) {
-    this.duration = props.duration !== undefined ? props.duration : 5000;
+    this.duration = props.duration !== undefined ? props.duration : DEFAULT_TIMEOUT;
```
ege0zcan, line 33: nit: `props duration ?? DEFAULT_TIMEOUT` seems easier unless I am missing some logic here

tverney: Yup, I did not changed the previous code, but sounds a better simplification. I'll include in the next rev.

src/static.ts
tverney: Sure, I've stick to the previous naming, but sounds better we tie to the tooltip. I'll include in the next rev.

src/components/navigation-tabs.ts
tverney: When its -1, we shouldn't set timeout per documentation.

ege0zcan: ah I see https://github.com/aws/mynah-ui/blob/main/docs/DATAMODEL.md#notificationprops

Though I don't think this overlay is being created as a notification so I am not sure if the `-1` applies here

ege0zcan (approved):

src/components/navigation-tabs.ts
tverney: The idea is keep the same behaviour that we have in notification (I know here we're using directly the Overlay): however, in API perspective, if user send -1, it would be infinite. We can include this in documentation, perhaps.

tverney:
@ege0zcan do you have writes to merge this one? If don't, I can ask someone else in the team. Thanks

Issue #81, opened 2024-07-18
Adding a scroll bar in Q chat


manodnyab:
## Description
While the scrolling functionality exists in Q chat panel, can we also add a visual component aka scroll bar to indicate the position in the chat panel?

32teeth:
This one should be trivial

this is why there is no scroll bar
```
    ::-webkit-scrollbar {
        width: 2px;
        height: 2px;
        opacity: 0.25;
        &:horizontal {
            width: 0px;
            height: 0px;
        }
    }    
```

PR #83, opened 2024-07-19
fix: #81 scrollbar

32teeth:
## Solution

Everything that needs / requires a scrollbar predicated by overflow conditions now has a scrollbar
* The gutter is `unset` meaning, it won't be there and be ugly.
* The draggable part is colored using css color mix blended in with white hs; `hsl(0, 100%, 100%)` @ `90%` which stays within the theme, and considerable enough for visual distinction

```
* {
  scrollbar-gutter: unset;
  scrollbar-color: color-mix(
    in hsl,
    var(--mynah-color-text-default),
    hsl(0, 100%, 100%) 90%)
    transparent;
}
```
Screenshot 2024-07-19 at 8 03 03 AM

## License

### DIFF
--- a/example/src/config.ts
+++ b/example/src/config.ts
@@ -4,8 +4,8 @@ import {
 import { defaultFollowUps } from './samples/sample-data';
 import { Commands } from './commands';
 import { MynahUITabStoreTab, QuickActionCommandGroup } from '../../dist/static';
-export const WelcomeMessage = `Hi, this is \`MynahUI\` and it is a **data and event driven** web based chat interface library and it is independent from any framework like react or vue etc. 
-In this example web app which uses mynah-ui as its renderer, we're simulating its capabilities with some static content with an IDE look&feel. 
+export const WelcomeMessage = `Hi, this is \`MynahUI\` and it is a **data and event driven** web based chat interface library and it is independent from any framework like react or vue etc.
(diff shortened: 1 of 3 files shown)


volodkevych:
Is there a reason to have scrollbar visible all the time? Where are requirements coming from? Should it be configurable? For example, Google Chrome has the same scrollbar style as MynahUI currently and I like it, tbh.

32teeth:
> Is there a reason to have the scrollbar visible all the time? Where are requirements coming from? Should it be configurable? For example, Google Chrome has the same scrollbar style as MynahUI currently, and I like it, to be honest.

The scrollbar by default never shows up for me, hence the change. The appearing/disappearing scrollbar will also be set by the user's OS preferences.

From an accessibility perspective, having the scrollbar always visible can benefit users with motor impairments or those who rely on visual cues for navigation. Ensuring that UI elements like scrollbars are consistently present can enhance usability for all users, making it easier to interact with content regardless of their individual needs or preferences. Thus, providing a configurable option allows us to cater to diverse accessibility requirements.

volodkevych (approved):

dogusata:
@32teeth this PR causes lots of issues related with the scrollable areas. 
For example the cursor position problem on prompt input come back. Additionally words are breaking in the middle which is more important than the scroll visibility at all. 

My suggestion is: 
If it is working, please do a feature creation task and wisely test it. Breaking the words in the middle of a chat item body is not an experience we want.

Issue #87, opened 2024-08-06
Command selection fails when switching focus or copy pasting a prompt


willyyhuang:
## Steps to reproduce the issue

1. Go to the mynah UI demo page: https://aws.github.io/mynah-ui/
2. type `/card-with-a-form` without pressing enter
3. Click on anywhere other than input box to lose focus
4. Click the send button
5. Copy `/card-with-a-form` 
6. Paste `/card-with-a-form` at the input box and press enter
7. Observe that "Amazon Q is generating..." is present

## Expected behavior
Should execute quick action when these 2 use cases are encountered

## System details

- OS: MacOS
- Tested enviroment:
    - [✅] Web (Demo app)
    - [✅] VSCode Amazon Q Chat
    - [✅] JetBrains IntelliJ Amazon Q Chat
- Enviroment extension version (if applicable): 
- [❌] Enviroment is remote

PR #88, opened 2024-08-06
fix: prompt quick action handling edge cases

willyyhuang:
## Problem

See #87 

## Solution

- Added a handling logic for prompt that starts with one of the quick action commands. It will now execute the first match when user lost focus or copy paste the prompt causing the quick action to not be selected.

- Before

https://github.com/user-attachments/assets/9851e86c-0048-4a28-8645-6c732682f084

- After

https://github.com/user-attachments/assets/b5ed474d-03d9-4eff-afed-4d0d9e9365a8

## License

### DIFF
--- a/src/components/chat-item/chat-prompt-input.ts
+++ b/src/components/chat-item/chat-prompt-input.ts
@@ -458,10 +458,28 @@ export class ChatPromptInput {
   };
 
   private readonly sendPrompt = (): void => {
+    const quickPickItems = MynahUITabsStore.getInstance().getTabDataStore(this.props.tabId).getValue('quickActionCommands') as QuickActionCommandGroup[];
     const currentInputValue = this.promptTextInput.getTextInputValue();
     if (currentInputValue.trim() !== '' || this.selectedCommand.trim() !== '') {
+      let selectedCommand = this.selectedCommand;
+
+      // Catching cases where user could send a prompt with quick action command but the command is not be selected correctly
+      if (selectedCommand === '') {
+        for (const quickPickItem of quickPickItems) {
+          if (selectedCommand !== '') break;
+          const matchedCommand = quickPickItem.commands.find((item) => currentInputValue.startsWith(item.command));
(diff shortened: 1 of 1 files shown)


src/components/chat-item/chat-prompt-input.ts
```diff
@@ -458,24 +458,43 @@ export class ChatPromptInput {
...
+    const quickPickItems = MynahUITabsStore.getInstance().getTabDataStore(this.props.tabId).getValue('quickActionCommands') as QuickActionCommandGroup[];
     const currentInputValue = this.promptTextInput.getTextInputValue();
     if (currentInputValue.trim() !== '' || this.selectedCommand.trim() !== '') {
+      let selectedCommand = this.selectedCommand;
+
+      // Catching cases where user could send a prompt with quick action command but the command is not be selected correctly
+      if (selectedCommand === '') {
+        for (const quickPickItem of quickPickItems) {
+          if (selectedCommand !== '') break;
+          const matchedCommand = quickPickItem.commands.find((item) => currentInputValue.startsWith(item.command));
```
tverney, line 470: if the the commend starts with a space " /dev" it still be activated?

willyyhuang: nope, it won't

tverney: Do we want to trigger that if this happens? We can potentially trim the space before command so it triggers in the same way.

willyyhuang: I don't think that's an intended behaviour, if you try right now, ` /` does not trigger quick action.

tverney: Yes, I know that currently does not trigger quick action. 

However, I don't see another intent for the user rather than open a new tab with the command even if add mistakenly a space before. I'm wondering if we shouldn't fix this also as part of the customer feedback.

tverney (approved):

dogusata (approved):

PR #91, opened 2024-08-07
fix: header in card not wrapping

willyyhuang:
## Problem

Header does not seem to wrap well with default `word-wrap: normal`

## Solution

before: 
image

after:
image

## License

### DIFF
--- a/src/styles/components/card/_card-body.scss
+++ b/src/styles/components/card/_card-body.scss
@@ -53,6 +53,9 @@
     h4,
     h5,
     h6 {
+      word-break: normal;
+      overflow-wrap: break-word;
+      line-height: calc(var(--mynah-line-height) * 1.25);
       margin-block-start: 0.75em;
       margin-block-end: 0.75em;
       padding-bottom: 1px !important;


src/styles/components/card/_card-body.scss
```diff
@@ -53,6 +53,7 @@
     h4,
     h5,
     h6 {
+      word-wrap: break-word;
```
dogusata, line 56: For any kind of situations, we're avoiding breaking words from the middle. This is a solution for the texts without spaces however it will also cause normal texts to break in the middle of the words. To avoid this, can we please apply the below solution? (including the line-height checks)

```css
word-break: normal;
overflow-wrap: break-word;
line-height: calc(var(--mynah-line-height) * 1.25);
```

dogusata (changes requested):

src/styles/components/card/_card-body.scss
willyyhuang: Resolved, thank you!

dogusata (approved):
🚀

PR #95, opened 2024-08-08
Fix conflict of @ context selections and user intended annotations start with @

dogusata:
## Problem
It is not possible to write any text which starts with @ character if they are not in the available contexts list. And if it is in the context list, it works as a context instead of user intended text.

## Solution
- Added checks to avoid removal of the word if it starts with @ character and not in the available context selection list
- Added checks to send only the words start with @ character in the available context items list as part of the prompt object
- Added checks to paint only the words start with @ character in the available context items list, otherwise keep as normal texts

Here's how it works:
<video width="500" height="500" src='https://github.com/user-attachments/assets/68ed9202-7110-4c9f-bc28-8b988ab9182e' />

NOTE: Even though the word is partially matching with one of the context items available, users are able to hit `esc` and proceed with their own texts. For example: after writing `@cod` if user directly hits space, it will automatically select `@code` as a context item. But if user hits escape and close the selector overlay, they can leave it as `@cod`.

## License

### DIFF
--- a/src/components/chat-item/chat-prompt-input.ts
+++ b/src/components/chat-item/chat-prompt-input.ts
@@ -48,10 +48,13 @@ export class ChatPromptInput {
         this.promptTextInputCommand.setCommand('');
       }
     });
+    const quickPickContextItems = (MynahUITabsStore.getInstance().getTabDataStore(this.props.tabId).getValue('contextCommands') as QuickActionCommandGroup[]) ?? [];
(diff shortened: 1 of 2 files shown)


senorflor:
Nice, thanks for addressing so quickly, and for generalizing the expected behaviors from the issue into a more coherent feature set!

ege0zcan (approved):

Issue #98, opened 2024-08-20
Bug report: Copy paste from MynahUI should not contain unnecessary line breaks


leigaol:
## Problem

This is a mirror of issue https://github.com/aws/aws-toolkit-vscode/issues/4991. 

## Steps to reproduce the issue

Resize chat window, then copy from chat window. The copied text contains line breaks that were added by MynahUI which does not exists in the original string. 

## System details

- OS:
- Tested enviroment:
    - [x] Web (Demo app)
    - [x] VSCode Amazon Q Chat
    - [x] JetBrains IntelliJ Amazon Q Chat
- Enviroment extension version (if applicable): 
- [-] Enviroment is remote

PR #100, opened 2024-08-21
Fix copy/paste for inline code

volodkevych:
## Problem

If you copy from the Mynah card with inline code and paste it into another place, it will contain new line breaks around inline code text.
The problem is described in more details in related issue: https://github.com/aws/mynah-ui/issues/98

## Solution

Avoid using block elements for inline code.

Before:
Screenshot 2024-08-21 at 18 12 28
After:
Screenshot 2024-08-21 at 18 12 44

## License

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@aws/mynah-ui",
-  "version": "4.15.7",
+  "version": "4.15.8",
   "lockfileVersion": 3,
   "requires": true,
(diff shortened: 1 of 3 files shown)


viktorsaws (approved):

volodkevych:
This change was verified in web, but still needs to be verified in IDEs.

dogusata (dismissed):

dogusata (approved):

PR #112, opened 2024-09-30
Pass userIntent from card to app on copy and code block action

bweedop:
## Problem

A chat response does not have information about the user's intent. User interactions with response, therefore, do not have information about the user's intent. The lack of information makes it difficult to track customer satisfaction with chat responses related to specific user intents.

## Solution

This PR adds the ability to provide a user's intent (if any) to chat items. It also adds the functionality needed to pass the user's intent from the chat message to the connected app (Amazon Q).

## License


dogusata:
So if i'm not mistaken, we're not doing anything with the `userIntent` on the MynahUI side. As I understand, It comes with the data, and without being used and/or manipulated and/or filled on the MynahUI side. It comes with the chat item data and goes back with the code block actions.
If it is the case, our main concern here is to make `mynah-ui` also a data transportation tool. If we're not using it on the mynah-ui side like visualizing something or to using it in some kind of logic, we do not receive or pass any data through mynah-ui. Since you have `messageId`, you can hold this information in the consumer app (VSCode toolkit, JetBrains toolkit etc.). Handling data depending on the messageId should be done on the consumer side. Like:

```typescript
const userIntentsOrAnyOtherData = {
'messageId': {...}
};

const mynahUI =  new MynahUI({
...
onCopyCodeToClipboard: (tabId, messageId, ...) => {
   console.log(userIntentsOrAnyOtherData[messageId]);
}
...
});
```

bweedop:
The reason I am asking is because we are trying to get acceptance metrics when a user clicks `copy` or `insert` on a Q response message. More specifically, we want to know how often users are clicking on those buttons for the different `userIntent` types. To achieve this, we want to associate the `userIntent` of a user's prompt with the response from Q. Having the `userIntent` on the response from Q will then allow us to emit `interactWithMessage` telemetry events with `userIntent`.

There are two ways I see we can accomplish this:

1) Pass the `userIntent` with the chat message like I did here. This would effectively store the `userIntent` in the chat component and then could be passed back to the consuming app (Amazon Q) when a user clicks on buttons like copy or insert. 

2) Store a history of prompts and responses (along with metadata like `userIntent`) in the toolkit as you suggest. The store, however, does not exist AFAIK.

Option 1 would allow for short-term mitigation. I agree though that, option 2, is preferable long-term to avoid using a UI library for data transfer back and forth to the consuming app.

PR #121, opened 2024-10-09
Jurredr/character count overlay

Jurredr:
## Problem
The character limit on the input always shows, but most users won't hit this limit, so this just takes up screen estate unnecessarily.

## Solution
We hide the character counter until the user approaches it. By default it will show at 3500 characters, but this is configurable in the config.

## License


dogusata (approved):
Awesome work!

PR #125, opened 2024-10-09
Dogusata/add progress bar to prompt input

dogusata:
## This PR includes
- Added progress indicator component
- Added progress options to prompt input field
- Fixed right shift bug of send button if the given text is too long without breaks
- Reverting the readable css custom properties

## License


src/components/chat-item/prompt-input/prompt-progress.ts
```diff
@@ -0,0 +1,46 @@
...
+        MynahUIGlobalEvents.getInstance().dispatch(MynahEventNames.PROMPT_PROGRESS_ACTION_CLICK, {
+          tabId: this.props.tabId,
+          actionId: action.id,
+          actionText: action.text
+        });
+      }
+    });
+    this.render = this.progressIndicator.render;
+
+    MynahUITabsStore.getInstance().getTabDataStore(this.props.tabId).subscribe('promptInputProgress', (progressData) => {
```
Jurredr, line 37: Don't we want to extract this 'promptInputProgress' to a static file like we do with MynahEventNames for example?

src/main.ts
```diff
@@ -404,6 +411,19 @@ export class MynahUI {
       }
     });
 
+    MynahUIGlobalEvents.getInstance().addListener(MynahEventNames.PROMPT_PROGRESS_ACTION_CLICK, (data: {
+      tabId: string;
+      actionId: string;
+      actionText?: string;
+    }) => {
+      if (this.props.onChatPromptProgressActionButtonClicked !== undefined) {
```
Jurredr, line 419: Is it not better to use != instead of !== here?

src/styles/components/_progress.scss
```diff
@@ -0,0 +1,81 @@
...
+    position: relative;
+    border-radius: inherit;
+    width: 100%;
+    display: block;
+    opacity: 1;
+    transition: var(--mynah-bottom-panel-transition);
+    &.no-content,
+    &.no-content * {
+        opacity: 0;
+        // display: none!important;
```
Jurredr, line 12: Probably should get rid of this line just for the sake of cleaning up

Jurredr (commented):
Nice work! Left a couple of small comments, nothing major. :)

example/src/main.ts
```diff
@@ -571,6 +589,25 @@ Use \`@\` to mention a file, folder, or method.`
...
+                valueText: '',
+                value: 100,
+                actions: []
+            }
+          });
+
+          mynahUI.updateStore(tabId, {
+            loadingChat: false,
+          });
+          setTimeout(()=>{
```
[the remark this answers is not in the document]
dogusata: this is the example app, almost everything is to simulate server responses.

src/components/chat-item/chat-item-buttons.ts
```diff
@@ -11,11 +11,11 @@ import { Icon } from '../icon';
 import { ChatItemFormItemsWrapper } from './chat-item-form-items';
 
 export interface ChatItemButtonsWrapperProps {
-  tabId: string;
+  tabId?: string;
```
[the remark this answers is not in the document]
dogusata: To make them usable also outside the tab-content area. For example feedback form too.

src/components/chat-item/prompt-input/prompt-progress.ts
dogusata: No, they're already coming automatically from the list of the keys in the tab-store. So they both support IntelliSense and typescript definitions already.

Jurredr (approved):

Issue #130, opened 2024-10-11
Make code blocks wrap the text on narrow screens


Jurredr:
## Problem
For readability of the code blocks in narrow screen conditions, it is better to wrap the code block texts since horizontal scrolling will cause too much movement.

## Expected behavior
We need to make the code blocks wrap the texts if the screen is narrower than 300px. If wider than that, it will remain same (horizontal scroll).

PR #131, opened 2024-10-11
#130: Code blocks now wrap text on screens narrower than 300px

Jurredr:
## Problem
For readability of the code blocks in narrow screen conditions, it is better to wrap the code block texts since horizontal scrolling will cause too much movement.

## Solution
Code blocks now take on a `white-space` of `pre-wrap` on screens narrower than 300px, ensuring the code is wrapped instead of overflowing with a scroll bar.

## License

### DIFF
--- a/src/styles/components/_syntax-highlighter.scss
+++ b/src/styles/components/_syntax-highlighter.scss
@@ -150,6 +150,10 @@ pre > code.diff-highlight .token.inserted:not(.prefix) {
             filter: brightness(0.95);
             white-space: pre;
             background-color: inherit;
+
+            @media (max-width: 300px) {
+                white-space: pre-wrap;
+            }
         }
 
         > code,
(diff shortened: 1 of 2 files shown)


src/styles/components/_syntax-highlighter.scss
```diff
@@ -161,6 +165,11 @@ pre > code.diff-highlight .token.inserted:not(.prefix) {
             word-wrap: normal;
             tab-size: 4;
             hyphens: none;
+
+            @media (max-width: 300px) {
```
dogusata, line 169: Please define this `300px` in css custom properties to allow consumer to update it without updating the whole `mynah-ui`. This 300px might not be applicable for every case. 

Additionally, we're trying avoid using static values for all cases. If it is not mandatory.

dogusata (changes requested):

dogusata (approved):

PR #140, opened 2024-10-16
fix: add data-label property to buttons

ivikash:
Add data-label attributes DOM. These data attributes can then be used by IDEs to define custom CSS, based on the button text.

Screenshot 2024-10-16 at 3 13 42 PM
Screenshot 2024-10-16 at 3 14 25 PM

## License

### DIFF
--- a/src/components/__test__/button.spec.ts
+++ b/src/components/__test__/button.spec.ts
@@ -79,4 +79,20 @@ describe('button', () => {
     testButtonElement.dispatchEvent(new Event('mouseenter'));
     expect(mockMouseOverHandler).toHaveBeenCalledTimes(1);
   });
+
+  it('data-label attribute', () => {
+    const mockOnClickHandler = jest.fn();
+    const testButton = new Button({
+      label: 'Test button',
+      onClick: mockOnClickHandler,
+    });
+    const testButton2 = new Button({
+      onClick: mockOnClickHandler,
+    });
+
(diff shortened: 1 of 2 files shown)


src/components/button.ts
```diff
@@ -63,6 +63,7 @@ class ButtonInternal extends ButtonAbstract {
       testId: props.testId,
       attributes: {
         ...(props.disabled === true ? { disabled: 'disabled' } : {}),
+        ...(Boolean(props.label) && typeof props.label === 'string' ? { 'data-label': props.label } : {}),
```
Jurredr, line 66: First check is unnecessary, type === 'string' will catch that it's not null

src/components/__test__/button.spec.ts
```diff
@@ -79,4 +79,20 @@ describe('button', () => {
     testButtonElement.dispatchEvent(new Event('mouseenter'));
     expect(mockMouseOverHandler).toHaveBeenCalledTimes(1);
   });
+
+  it('data-label attribute', () => {
```
Jurredr, line 83: Small nit, perhaps add a quick check to show that testButton2 will also get the attribute once `updateLabel` is called on it, even though it didn't have a label to start out with

dogusata:
We're not allowing custom css for buttons or any other component. In case different type of button styles needed, they should be added and can be picked up by using data properties. Currently, this is not a component library, and we won't allow overriding elements. This can be done when the component's are being exposed as a component library in a different repo.

PR #141, opened 2024-10-17
removed eventId check from selectTab

dogusata:
## Problem
It is not possible to switch to a tab without user intent from the **inside** of mynah-ui. However, consumer can already have a user intent (for example by showing a notification on IDE and letting the user click it), In this case, even though there is a user intent, it will mismatch with the one on the mynah-ui side. 
## Solution
Until finding a proper solution to between the consumer and the mynah-ui, we're skipping the eventId check,

## License

### DIFF
--- a/.github/workflows/beta.yml
+++ b/.github/workflows/beta.yml
@@ -14,9 +14,7 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
           scope: '@aws'
       - run: npm ci
-      - run: npm run lint
       - run: npm run build
-      - run: npm test
       - run: npm publish --tag beta --access public
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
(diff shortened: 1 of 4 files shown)


.github/workflows/beta.yml
```diff
@@ -14,9 +14,7 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
           scope: '@aws'
       - run: npm ci
-      - run: npm run lint
       - run: npm run build
-      - run: npm test
```
Jurredr, line 19: No linting and tests on beta?

dogusata: No, not required, I want to keep the beta release strategy as fast as possible and consumer should be aware that it is an untested beta version. We may still improve it with less tests and checks but for now not necessary.

Jurredr (approved):

PR #144, opened 2024-10-18
fix: only re-focus after a response stream if the window is active

jpinkney-aws:
## Problem
In VSCode if you:
    - write a message in chat
    - focus somewhere else in the editor outside of chat
    - wait for the response stream to finish
it will automatically focus the chat prompt

## Solution
Only refocus the chat prompt if the document is currently focused

## License

### DIFF
--- a/src/components/chat-item/prompt-input/prompt-text-input.ts
+++ b/src/components/chat-item/prompt-input/prompt-text-input.ts
@@ -115,7 +115,7 @@ export class PromptTextInput {
       } else {
         // Enable the input field and focus on it
         this.promptTextInput.removeAttribute('disabled');
-        if (Config.getInstance().config.autoFocus) {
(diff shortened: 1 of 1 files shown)


dogusata (approved):
Great!

dogusata:
To clarify the problem related with the solution a bit more in detail:
In VSCode, when user has the focus to the Q panel, we want to autofocus to the prompt input field after a streaming answer is ended if the `autoFocus` config is set to true. 
However, it should only do it when the current focus is inside the Q panel window. Otherwise, it will grab the focus from the editor to the Q panel which might cause an issue since user may write to their editor window or focus to another panel while waiting an answer. And while they are typing in their editor window sudden switch to the Q panel may cause the text written by the user went into the Q prompt field instead of their editor file.

PR #146, opened 2024-10-22
Dogusata/add capability card

Jurredr:
## Description
Added a `informationCard` option for chat items, which can (for example) be used to show off particular features.

image
image

## License


src/components/chat-item/chat-item-card.ts
```diff
@@ -299,6 +301,101 @@ export class ChatItemCard {
       this.card?.render.insertChild('beforeend', this.fileTreeWrapper.render);
     }
 
+    /**
```
dogusata, line 304: Comment doesn't match with the content below.

src/components/chat-item/chat-item-card.ts
```diff
@@ -299,6 +301,101 @@ export class ChatItemCard {
...
+     * Generate file tree if available
+     */
+    if (this.informationCardWrapper != null) {
```
dogusata, line 307: I've thought about it several times, but if we want to use it in places other than the chat-item-card, it would be handy to have this component as a separate component. 

Let's move it to a different file and just call the component here (like the `ChatItemButtons`). 
Together with the `status` block.

src/styles/_variables.scss
```diff
@@ -43,10 +43,26 @@
     --mynah-color-status-warning: #b2911c;
     --mynah-color-status-error: #d91515;
 
+    --mynah-color-status-info-bg: rgba(9, 113, 211, 0.15);
```
dogusata, line 46: Instead of duplicating colors with different opacities, it might be better to use pseudo elements for the background and play with its opacity. 

In case if you have to use a dedicated color (not applicable for the informationBlocks status section), it means that they don't need to be exposed as themeable items  and you can generate them through scss functions.

src/components/chat-item/chat-item-card.ts
Jurredr: Fixed

dogusata (changes requested):

example/src/samples/sample-data.ts
```diff
@@ -712,3 +713,70 @@ export const exampleDownloadFile: ChatItem = {
...
+        title: 'Information card',
+        description: 'With a description below the title.',
+        icon: MynahIcons.BUG,
+        content: {
+            body: sampleList2 as string
+        },
+    },
+};
+
+export const exampleInformationCardWarning: ChatItem = {
```
dogusata, line 730: nit: Maybe just a function to send changing parameters instead of creating multiple ones.

src/components/chat-item/chat-item-information-card.ts
```diff
@@ -0,0 +1,109 @@
+import { DomBuilder, ExtendedHTMLElement } from '../../helper/dom';
+import { ChatItemContent, ChatItemType, Status } from '../../static';
+import { Icon, MynahIcons } from '../icon';
+import { ChatItemCard } from './chat-item-card';
+
+export interface ChatItemInformationCardProps {
+  tabId: string;
+  messageId: string | undefined;
+  informationCard: {
```
dogusata, line 9: why not using and extending the `ChatItemContent` type?
Also, having the `testId` property might be useful here as well.

dogusata (approved):
Approving with 2 nit followups.

PR #152, opened 2024-10-31
Autofix: URLs barely visible on VSCode light theme

lattaai1:
I've updated the _variables.scss file to adjust the color of text links in the light theme. This change will make URLs more visible when using the VSCode light theme. The modification involves changing the value of --mynah-color-text-link to use a darker shade when the light theme is active. 
> [!CAUTION]  
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission


dogusata:
NO.

dogusata:
> This bug was fixed for free by Latta AI

And sure it is not fixed. What about dark themes?

PR #155, opened 2024-11-05
dogusata/Add Welcome screen additions

dogusata:
## Problem
New welcome screens needs to be supported

## Solution
- Added `compactMode` to tab store
- Added `tabHeaderDetails` to tab store
- Added `compactMode` to tab store
- Added `tabBackground` to tab store
- Added `promptInputLabel` to tab store (to show label on top of the prompt input)
- Added `promptInputText` to tab store (to set the text of prompt input dynamically, also to open quick actions picker)
- Added `promptInputVisible` to tab store (to show/hide prompt input field dynamically for each tab unlike the global config)
- Added `hoverEffect` to ChatItem (which gives a nice upwards movement to the card when hovered)
- Added `flash` for `buttons` and `codeBlockActions` items (which flashes the button when hovered to container card)
- Added `TitleDescriptionWithIcon` component
- Added `GradientBackground` component (which grabs colors from the css custom properties)
- Added `main` and `clear` color options to buttons
- Added `ROCKET` icon
- Added `onTabbedContentTabChange` event property to main class
- Added E2E tests and snapshots
- Updated `Q` icon with colored one
- Updated `tabbedCard` styles
- Updated `informationCard` styles (with new `TitleDescriptionWithIcon` component)
- Updated test snapshots of information card and tabbed cards.
- Fixed tooltips for disabled state of the buttons
- Fixed footer layout positioning

## Tests

## License


src/styles/_variables.scss
```diff
@@ -52,6 +52,8 @@
     --mynah-bg-gradient-mid: #538bf7;
     --mynah-bg-gradient-start: #301673;
     --mynah-bg-gradient-end: #6e3bf1;
+    --mynah-bg-gradient-next: #da8af7;
```
Jurredr, line 55: What are these used for?

src/styles/components/chat/_chat-item-card-information-card.scss
```diff
@@ -13,118 +13,85 @@
...
-            }
-        }
-    }
-
-    &:has(.mynah-chat-item-information-card-footer) {
         > .mynah-chat-item-information-card-main {
             border-bottom-left-radius: 0;
             border-bottom-right-radius: 0;
-            border-bottom: 0;
+            border-bottom: none !important;
```
Jurredr, line 94: Any difference between `0` and `none`?

src/styles/components/chat/_chat-item-tree-view.scss
```diff
@@ -95,6 +96,18 @@
             min-height: auto;
             height: auto;
             margin: 0;
+            &.mynah-chat-item-tree-view-not-clickable {
+                cursor: default;
+                &:active,
+                &:focus-visible,
+                &:hover {
+                    filter: brightness(0.95);
+                    &:after {
```
Jurredr, line 105: What does this after within the hover do exactly?

src/styles/components/chat/_chat-wrapper.scss
```diff
@@ -47,6 +49,106 @@
...
+        opacity: 0;
+        > .mynah-ui-title-description-icon-icon > .mynah-ui-icon {
+            font-size: var(--mynah-font-size-large);
+        }
+        > .mynah-ui-title-description-icon-title {
+            font-weight: bold;
+        }
+        > .mynah-ui-title-description-icon-title,
+        > .mynah-ui-title-description-icon-description {
+            font-size: calc(2 * var(--mynah-font-size-xxsmall));
```
Jurredr, line 139: Why use a 2x of a variable here? Wouldn't twice the size of `xxs`, indicate something like `sm` (with a variable)?

src/components/chat-item/chat-wrapper.ts
```diff
@@ -28,15 +30,27 @@ export class ChatWrapper {
...
   private readonly promptStickyCard: ExtendedHTMLElement;
+  private tabHeaderDetails: ExtendedHTMLElement;
+  private tabModeSwitchTimeout: ReturnType<typeof setTimeout> | null;
   private lastStreamingChatItemCard: ChatItemCard | null;
   private lastStreamingChatItemMessageId: string | null;
   private allRenderedChatItems: Record<string, ChatItemCard> = {};
   render: ExtendedHTMLElement;
   constructor (props: ChatWrapperProps) {
     this.props = props;
-    MynahUITabsStore.getInstance().getTabDataStore(this.props.tabId).subscribe('chatItems', (chatItems: ChatItem[]) => {
```
Jurredr, line 39: I'm confused, what is the difference between `subscribe()` and `addListenerToDataStore`? Or is this a renamed function?

example/src/main.ts
```diff
@@ -239,7 +281,35 @@ export const createMynahUI = (initialData?: MynahUIDataModel): MynahUI => {
...
+          tabHeaderDetails: null,
+          compactMode: false,
+          tabBackground: false,
+          promptInputText: '/dev',
+          promptInputLabel: null,
+          chatItems: []
+        });
+      }
+      if (action.id === 'explore') {
+        // mynahUI.updateStore(tabId, {chatItems: []});
```
Jurredr, line 302: remove commented line

src/components/button.ts
```diff
@@ -27,9 +34,10 @@ export interface ButtonProps {
   disabled?: boolean;
   primary?: boolean;
   border?: boolean;
-  status?: 'primary' | 'info' | 'success' | 'warning' | 'error';
-  additionalEvents?: Record<string, (event?: any) => any>;
+  status?: 'main' | 'primary' | 'info' | 'success' | 'warning' | 'error' | 'clear';
```
Jurredr, line 37: `main` vs `primary` does not seem to make much sense in general status hierarchy, bit ambiguous?

src/styles/components/_button.scss
```diff
@@ -60,6 +58,22 @@
         }
         box-shadow: none;
     }
+    &.status-clear {
+        &,
```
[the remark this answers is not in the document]
dogusata: self and the next selector.

src/styles/components/_main-container.scss
```diff
@@ -33,6 +33,8 @@ body.mynah-ui-root {
         max-width: inherit;
         display: block;
         height: inherit;
+        // Flexbox fix
```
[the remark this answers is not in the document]
dogusata: It actually fixes the `flex-grow` and `flex-shrink` according to the parent height. When there is no `overflow: hidden` set.

src/styles/components/chat/_chat-item-card-information-card.scss
dogusata: Correct way to do it is none. Depending on GPU, it can still render but without a width. Both will not be visible, however the none will skip the rendering part completely.

src/styles/components/chat/_chat-item-tree-view.scss
dogusata: That is the flashing cover item, it's main definition is at above. This is just for the animation (which comes over the button) when you hover.

src/styles/components/chat/_chat-items-container.scss
```diff
@@ -1,7 +1,8 @@
 > .mynah-chat-items-container {
     position: relative;
     display: flex;
-    flex: 1 1 0;
```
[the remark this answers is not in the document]
dogusata: It is being controlled through the wrapper class now. And also, the given values above will cause the footer to exceed from the bottom as well.

src/styles/components/chat/_chat-wrapper.scss
dogusata: xxs X 2 is not equal to sm. And even the largest one is not enough big.

`--mynah-font-size-xxsmall: 0.8rem;` and we need `1.6rem` for the title and description part

src/styles/_variables.scss
dogusata: These are 2 additional values for the background, we need to validate with the UX team first, but i also don't want to add more colors for the gradient. Until we have a validation from the UX team, they'll stay.

src/components/chat-item/chat-wrapper.ts
dogusata: They're doing the exact thing. However keep it consistent through the whole usage, preferred way is `addListenerToDataStore`. Actually `subscribe` should be private.

src/components/button.ts
dogusata: Definitely yes, but it will be a task for a major version change to v5.0.0. We'll have types for the buttons, and status will be replaced with `status-info`, `status-primary` etc. However we don't need to make a major and backwards incompatible change for the consumers.

src/styles/components/chat/_chat-wrapper.scss
Jurredr: Does it not make sense to create a new (very large) sizing variable, then?

src/components/chat-item/chat-wrapper.ts
Jurredr: So why do we have both if they do the exact same thing?

src/components/button.ts
Jurredr: But what is main here? The gradient?

src/components/chat-item/chat-wrapper.ts
dogusata: One of them is one level deep, the other one is directly on the tabStore object.

So: `tabStore. addListenerToDataStore` and `tabStore.getTabStoreData().subscribe`. 
For example If you already grabbed the tabStoreData before, you can use it directly.
```typescript
const storeData = MynahUITabsStore.getInstance().getTabDataStore(tabId);
// do some stuff with storeData
storeData.subscribe(...)
```

But if you directly need to listen a value change on the store:
```typescript
 MynahUITabsStore.getInstance().addListenerToDataStore(...)
```

src/components/button.ts
dogusata: Yes it is the gradient one. 
In v5 we'll arrange them under `type` and will be like:
```
type: 'main' | 'primary' | 'secondary' | 'clear';
// if type is "secondary"
status: Status;
```

we'll define them later

Jurredr: Does a type of 'gradient' not make more sense then? 'main' and 'primary' are literal synonyms.

dogusata: Not for this case, since it is gradient now, but maybe in the future we'll replace it with something else. So it is same with naming with colors. There cannot be a `blue` or `red` button for example, since our lib is themeable and those can change.

src/styles/components/chat/_chat-wrapper.scss
dogusata: It is there

dogusata: Just added

Jurredr (approved):

PR #157, opened 2024-11-07
Fix disappearing span tags inside code blocks

ege0zcan:
## Problem

The issue was reported in: https://github.com/aws/aws-toolkit-vscode/issues/5732

When there is a closing `span` tag inside a code block or inline code, it currently disappears and is not rendered. This is caused by old code in mynah-ui that would do some pre-processing on the html contents and replace certain span elements before they were rendered. This causes any span tag to be replaced and removed before they are rendered and this leads to the bug described above

See for example how the following input is rendered currently:

````
```
<span> span inside a codeblock </span>
```

`<span> span inside inline code </span>`

`````
Screenshot 2024-11-07 at 11 20 56

## Solution

To remove the old `keepHighlights` functionality. 

See an example card body that correctly renders the span tags after the fix: 

Screenshot 2024-11-07 at 11 24 00

And the test case from above with prompt input containing spans:
 
Screenshot 2024-11-07 at 11 44 08

## Tests
- [x] I have tested this change on JetBrains

## License

### DIFF
--- a/src/components/__test__/syntax-highlighter.spec.ts
+++ b/src/components/__test__/syntax-highlighter.spec.ts
@@ -5,7 +5,6 @@ describe('syntax-highlighter', () => {
     const testSyntaxHighlighter = new SyntaxHighlighter({
       codeStringWithMarkup: 'alert("hello");\n',
       language: 'js',
-      keepHighlights: true,
       block: true,
     });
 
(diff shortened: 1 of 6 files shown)


dogusata (dismissed):

Jurredr:
image
Small issue spotted in follow ups, @dogusata is on it.

Jurredr (approved):

Jurredr:
Screenshot 2024-11-07 at 14 57 58
Screenshot 2024-11-07 at 14 57 48
All fixed!

PR #164, opened 2024-11-11
163 quick action commands confirmation

Jurredr:
## Problem
Quick commands without parameters would instantly run when clicked through the menu.

## Solution
Instead of instantly running the command, we show a confirmation text in the prompt input, which can be confirmed by pressing enter and denied by unfocusing or pressing escape.

## Tests

## License


justinmk3:
Is there a purpose to these checklists?

image

Jurredr:
@justinmk3 We've mainly added that checklist to encourage contributors outside of the core team to think about actually testing larger (possibly breaking) changes in IDEs, as we've had some contributions come through which seemed fine at first but ended up not working properly in the actual IDE environments.

justinmk3:
Everyone just ignores it, so it's noise.

Jurredr:
I think it'd be better to add an alternative option stating something along the lines of "This PR doesn't necessarily require in-IDE testing".

src/components/chat-item/chat-prompt-input.ts
```diff
@@ -542,7 +555,11 @@ export class ChatPromptInput {
       this.promptTextInputCommand.setCommand(this.selectedCommand);
       this.promptTextInput.updateTextInputPlaceholder(quickActionCommand.placeholder);
     } else {
-      this.sendPrompt();
+      this.promptTextInputCommand.setCommand(this.selectedCommand);
+      // TODO: get this from config
```
dogusata, line 559: You're already getting this from config, you can remove the TODO

dogusata:
> I think it'd be better to add an alternative option stating something along the lines of "This PR doesn't necessarily require in-IDE testing". It's not being ignored in every PR, and the main reason for adding it was more so aimed at preventing external contributors from not being mindful about testing. Currently, we do not really have active external contributors.

No, we're not gonna add such kind of option and this PR definitely requires a test on vscode and jetbrains. Since it is customer facing, it is best to see them also in VSCode and IntelliJ before merging. 

As an additional note, let me mention the regression we've created related with the liveserver. Our tests were passing, everything was looking ok but we've it caused the installation process to broke for the VSCode (and also most likely the JetBrains).

Depending on the PR (like changing some text etc. or just adding a new test into e2e tests) we can decide it is required or not.

dogusata:
> Everyone just ignores it, so it's noise. Noise should be reduced.

We'll ask for a specific reason to ignore it in the PRs. We want to avoid any kind of regression as much as possible. And this checks shouldn't be skipped.

Again, as an example: to avoid regressions like `postinstall` `liveserver` addition.

dogusata (approved):
After its being checked for VSCode and JetBrains, it can be merged! Great work!

PR #180, opened 2024-11-22
Add up down arrow navigation in prompt field to access prompts in history

francescoopiccoli:
## Problem
Up and down keys inside the prompt input field should switch between previous prompts like in a terminal. To increase the UX and let the end user easily access to a previous prompt to modify/improve it, the approach used in terminals needs to be applied.

Once I receive validation for my implementation approach I will proceed with adding tests and documentation.

## Solution
Retrieving and identifying commands and code snippets from previous chat prompts, required adding some extra properties and methods that were not current available. 

## Tests

## License

### DIFF
--- a/example/src/main.ts
+++ b/example/src/main.ts
@@ -39,6 +39,7 @@ import {
   exampleConfirmationButtons,
 } from './samples/sample-data';
 import escapeHTML from 'escape-html';
+import unescapeHTML from 'unescape-html';
 import './styles/styles.scss';
 import { ThemeBuilder } from './theme-builder/theme-builder';
 import { Commands } from './commands';
@@ -229,6 +230,11 @@ export const createMynahUI = (initialData?: MynahUIDataModel): MynahUI => {
           promptInputLabel: null,
         });
       }
+      // reset navigation index when sending a new prompt
+      mynahUI.updateStore(tabId, {
+        navigationIndexChatPrompts: -1,
+      })
+
       onChatPrompt(tabId, prompt);
     },
     onStopChatResponse: (tabId: string) => {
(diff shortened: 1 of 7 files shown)


src/static.ts
```diff
@@ -311,6 +319,8 @@ export interface ChatPrompt {
   escapedPrompt?: string;
   command?: string;
   context?: string[];
+  attachmentContent?: string;
+  inputText?: string;
```
Jurredr, line 323: What is this in comparison to prompt?

src/static.ts
```diff
@@ -245,6 +250,9 @@ export interface TreeNodeDetails {
 
 export interface ChatItemContent {
   body?: string | null;
+  inputText?: string | null;
```
Jurredr, line 253: Just an idea; might be nice to bundle these 3 under something like `inputs: {}`

src/static.ts
```diff
@@ -126,6 +126,10 @@ export interface MynahUIDataModel {
    * Tab content header details, only visibile when showTabHeaderDetails is set to 'true'
    */
   tabHeaderDetails?: TabHeaderDetails | null;
+  /**
+   * The navigation index for the chat prompts list for the up/down arrow key navigation
+   */
+  navigationIndexChatPrompts?: number;
```
Jurredr, line 132: Maybe change to `upDownKeyNavigationIndex` for consistency with `UP_DOWN_ARROW_KEY_PRESS` naming

example/src/main.ts
```diff
@@ -387,6 +393,64 @@ export const createMynahUI = (initialData?: MynahUIDataModel): MynahUI => {
     onInfoLinkClick: (tabId, link, mouseEvent) => {
       Log(`Link inside prompt info field clicked: <b>${link}</b>`);
     },
+    upDownArrowKeyPress(tabId: string, direction: 'up' | 'down') {
```
Jurredr, line 396: Rename to "**on**UpDownArrowKeyPress"

example/src/main.ts
```diff
@@ -387,6 +393,64 @@ export const createMynahUI = (initialData?: MynahUIDataModel): MynahUI => {
...
       Log(`Link inside prompt info field clicked: <b>${link}</b>`);
     },
+    upDownArrowKeyPress(tabId: string, direction: 'up' | 'down') {  
+      // reset promptInputText and any possible attachment/commands from previous prompts
+      mynahUI.updateStore(tabId, {
+        promptInputText: '',
+      });
+      mynahUI.clearPrompt(tabId);
+
+      const tab = mynahUI.getAllTabs()[tabId];
```
Jurredr, line 403: rename to `currentTab` for clarity. This is usually our convention when we grab the current tab.

example/src/main.ts
```diff
@@ -387,6 +393,64 @@ export const createMynahUI = (initialData?: MynahUIDataModel): MynahUI => {
...
+  
+      if (!chatPrompts.length) { return; }
+
+      let navigationIndex = tab.store.navigationIndexChatPrompts === -1 ? chatPrompts.length : (tab.store.navigationIndexChatPrompts ?? chatPrompts.length);  
+    
+      if (direction === 'up') {
+          navigationIndex = Math.max(0, navigationIndex - 1);
+      } else if (direction === 'down') {
+          navigationIndex = Math.min(chatPrompts.length, navigationIndex + 1);
+      } else {
```
Jurredr, line 417: Redundant check, we already type check the input as `'up' | 'down'`

src/components/chat-item/chat-prompt-input.ts
```diff
@@ -656,4 +667,9 @@ export class ChatPromptInput {
       type
     });
   };
+
+  public readonly addCommand = (quickActionCommand: string): void => {
```
Jurredr, line 671: rename to `setCommand` for consistency

src/main.ts
```diff
@@ -246,6 +246,10 @@ export interface MynahUIProps {
     tabId: string,
     buttonId: string,
     eventId?: string) => void;
+  upDownArrowKeyPress?: (
+    tabId: string,
+    direction: 'up' | 'down',
+    eventId?: string) => void;
```
Jurredr, line 252: huh? what is `string)`, isn't there a parenthesis missing? (same for line 248)

src/static.ts
dogusata: It is not so preferable to add these to typedef, if we're not planning to send it back to consumer. Since consumer uses this and they might expect to have an input text and attachment content. Instead, create a private interface extending from ChatPrompt for your use case

src/main.ts
dogusata: where is the open? :)

francescoopiccoli: Not sure I got this. Maybe its just a formatting on github review? 
Here's what I see:
```
onFileActionClick?: (
    tabId: string,
    messageId: string,
    filePath: string,
    actionName: string,
    eventId?: string) => void;
 onTabBarButtonClick?: (
    tabId: string,
    buttonId: string,
    eventId?: string) => void;
 onUpDownArrowKeyPress?: (
    tabId: string,
    direction: 'up' | 'down',
    eventId?: string) => void;
```

src/static.ts
francescoopiccoli: > We should already have them in the keys list.

sorry I didnt get this, not sure I know what is the key list and what you mean with 'them'?

PR #186, opened 2024-11-27
Add up down arrow navigation in prompt field to access prompts in history

francescoopiccoli:
## Problem
Up and down keys inside the prompt input field should switch between previous prompts like in a terminal. To increase the UX and let the end user easily access to a previous prompt to modify/improve it, the approach used in terminals needs to be applied.
## Solution
Retrieving and identifying commands and code snippets from previous chat prompts, required adding some extra properties and methods that were not current available.

## Tests

## License

### DIFF
--- a/src/components/chat-item/chat-prompt-input.ts
+++ b/src/components/chat-item/chat-prompt-input.ts
@@ -34,6 +34,12 @@ export const INPUT_LENGTH_WARNING_THRESHOLD = (): number => {
 export interface ChatPromptInputProps {
   tabId: string;
 }
+
+interface UserPrompt {
+  inputText: string;
+  codeAttachment: string;
+}
+
 export class ChatPromptInput {
   render: ExtendedHTMLElement;
   private readonly props: ChatPromptInputProps;
@@ -54,6 +60,8 @@ export class ChatPromptInput {
   private quickPick: Overlay;
   private quickPickOpen: boolean = false;
   private selectedCommand: string = '';
+  private readonly userPromptHistory: UserPrompt[] = [];
+  private userPromptHistoryIndex: number = -1;
   constructor (props: ChatPromptInputProps) {
     this.props = props;
     this.promptTextInputCommand = new ChatPromptInputCommand({
@@ -250,6 +258,8 @@ export class ChatPromptInput {
   };
 
   private readonly handleInputKeydown = (e: KeyboardEvent): void => {
+    const navigationalKeys = [ KeyMap.ARROW_UP, KeyMap.ARROW_DOWN ] as string[];
+
     if (e.key === KeyMap.ESCAPE && this.render.hasClass('awaits-confirmation')) {
       this.promptTextInput.blur();
     }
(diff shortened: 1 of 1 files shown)


src/components/chat-item/chat-prompt-input.ts
```diff
@@ -312,10 +322,47 @@ export class ChatPromptInput {
...
+          if (typeof codeAttachment === 'string' && codeAttachment.trim().length > 0) {
+            codeAttachment = codeAttachment
+              .replace(/~~~~~~~~~~/, '')
+              .replace(/~~~~~~~~~~$/, '')
+              .trim();
+            this.promptAttachment.updateAttachment(codeAttachment, 'code');
+          }
+
+          const promptLength = this.promptTextInput.getTextInputValue().trim().length + (typeof codeAttachment === 'string' ? codeAttachment.length : 0);
+          this.promptTextInput.updateTextInputMaxLength(Math.max(MAX_USER_INPUT_THRESHOLD, (MAX_USER_INPUT() - promptLength)));
```
francescoopiccoli, line 354: To update the text input max length I consider both attachment and current input text

src/components/chat-item/chat-prompt-input.ts
```diff
@@ -632,6 +679,14 @@ export class ChatPromptInput {
         }
       };
       this.clearTextArea();
+
+      if (currentInputValue !== '') {
+        this.userPromptHistory.push({
+          inputText: currentInputValue,
+          codeAttachment: attachmentContent,
+        });
+      }
```
francescoopiccoli, line 688: I would avoid checking length of userPromptHistory and set a maximum for now

src/components/chat-item/chat-prompt-input.ts
```diff
@@ -312,10 +322,47 @@ export class ChatPromptInput {
...
         }
+      } else if (navigationalKeys.includes(e.key)) {
+        const direction = e.key === KeyMap.ARROW_UP ? 'up' : 'down';
+        this.clearTextArea();
+
+        if (this.userPromptHistoryIndex === -1) {
+          this.userPromptHistoryIndex = this.userPromptHistory.length;
+        }
+
+        if (direction === 'up') {
```
dogusata, line 333: Why not using the KeyMap.ARROW_UP or _DOWN directly instead? I don't see any reason to put it to one more variable again.

src/components/chat-item/chat-prompt-input.ts
dogusata: So the things you'll get from the input and the attachment were already being calculated before they went into the elements. This recalculation seems not necessary. Also the same for replacements, they're already being applied.

src/components/chat-item/chat-prompt-input.ts
```diff
@@ -312,10 +322,47 @@ export class ChatPromptInput {
...
+          this.userPromptHistoryIndex = Math.max(0, this.userPromptHistoryIndex - 1);
```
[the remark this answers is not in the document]
francescoopiccoli: I would prefer to keep it as it is to keep it consistent with the terminal arrow navigation the user might be familiar with and maintain a sense of order/direction between prompts: if i go back to top of the list when we reach the first one, was the top of the list actually an old prompt or the latest one, the user might navigate the prompts based on how recent they sent them

src/components/chat-item/chat-prompt-input.ts
francescoopiccoli: without replacement the code attachment will show '~~~~~~~~~~'

francescoopiccoli: With recalculation you mean updating via `updateTextInputMaxLength` and `updateAvailableCharactersIndicator` right?

src/components/chat-item/chat-prompt-input.ts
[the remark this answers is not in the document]
Jurredr: +1 on keeping it consistent with terminal as @francescoopiccoli proposes.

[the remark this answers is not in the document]
dogusata: I also found myself to hold up arrow to hit the oldest one easily. Agreed!👍🏻 🚀

dogusata:
Where are the chromium snapshots?

dogusata (approved):

PR #193, opened 2024-11-29
Fix tests and update image match failure threshold

natdurlik:
## Problem
Low `failureThreshold` value for the `toMatchImageSnapshot` method used in the e2e tests causes tests to pass even when significant differences in the screenshots are visible to human eye.

## Solution
Previously the `failureThreshold` was set to 0.1, meaning screenshots could differ by 10% and the image comparison would still pass. This caused issues for screenshots where only a small part changed. For example, one small chat answer element was accounting for around 5% difference in the whole chat tab. See the PR changes for examples of tests that were still passing despite containing significant differences.

Changed the threshold to 0.01, which means 1%. This could be updated to be even smaller if there are no drawbacks.

## Tests

## License


Jurredr:
Really nice, we couldn't reliably get it under 10% before without extreme flakiness. Have you ran the tests multiple times to confirm it passes frequently (every time?) with the new golden snapshots and 1% threshold?

dogusata:
This is very nice! Thanks for the effort. 
In addition to that, another great thing would be getting the snapshot of the target elements instead of the whole screen. Which will increase the ratio of the mismatch significantly.

dogusata (approved):

PR #197, opened 2024-12-05
Adjust footer text size

dogusata:
## Problem
The footer text size and the line-height (in general) is too much and when there is a long disclaimer text it takes too much screen real estate.
image

## Solution
The given font size to the footer was being overridden, made it !important to avoid overrides. Also, reduced the line-height in general as requested.

### Additionally
We have much more tests now and they are taking too much time and resources on local. Removing the `tests:e2e` step from the prepush to leave the test runs to github actions.

## Tests

## License

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,3 +1,2 @@
 npm run lint
 npm run format:check
-npm run tests:e2e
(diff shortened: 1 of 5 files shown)


Jurredr (approved):
+1 on the removal of test runs on pre-push, it's undoable with how long it takes now. Another reason for the current considerations of a new approach to e2e testing.

PR #200, opened 2024-12-09
Fix code marks trimming for different ides

francescoopiccoli:
## Problem
in our example mynah-ui we mark codeblock with \~~~~~~~~~~, different ides use \``` instead. This causes a small ui issue when navigating through prompts as the code blocks attached to previous prompts might not be trimmed correctly. More importantly, in this PR we make sure to remove the code marks ONLY if they appear at the very start of the codeAttachment, not the first occurrence. This issue could have potentially modified the user prompt if it contained an occurrence of  \~~~~~~~~~~ while the code block was marked via the \``` marks. In that case navigating through prompts would have caused an input like:
```
void main() {
  print("~~~~~~~~~~");
  print("Hello World");
  print("Hello World");
}
```

to be transformed to 
```
void main() {
  print("");
  print("Hello World");
  print("Hello World");
}
```

## Tests

## License

### DIFF
--- a/src/components/chat-item/chat-prompt-input.ts
+++ b/src/components/chat-item/chat-prompt-input.ts
@@ -355,11 +355,20 @@ export class ChatPromptInput {
           });
           codeAttachment = this.userPromptHistory[this.userPromptHistoryIndex].codeAttachment ?? '';
         }
-        if (codeAttachment.trim().length > 0) {
-          codeAttachment = codeAttachment
-            .replace(/~~~~~~~~~~/, '')
-            .replace(/~~~~~~~~~~$/, '')
-            .trim();
+        codeAttachment = codeAttachment.trim();
+        if (codeAttachment.length > 0) {
+          // our example
+          if (codeAttachment.startsWith('~~~~~~~~~~') && codeAttachment.endsWith('~~~~~~~~~~')) {
+            codeAttachment = codeAttachment
+              .replace(/^~~~~~~~~~~/, '')
+              .replace(/~~~~~~~~~~$/, '')
+              .trim();
+          } else if (codeAttachment.startsWith('```') && codeAttachment.endsWith('```')) {
+            codeAttachment = codeAttachment
+              .replace(/^```/, '')
+              .replace(/```$/, '')
+              .trim();
+          }
           this.addAttachment(codeAttachment, 'code');
         }
       }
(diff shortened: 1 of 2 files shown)


src/components/chat-item/chat-prompt-input.ts
```diff
@@ -355,11 +355,20 @@ export class ChatPromptInput {
...
-            .trim();
+        codeAttachment = codeAttachment.trim();
+        if (codeAttachment.length > 0) {
+          // our example
+          if (codeAttachment.startsWith('~~~~~~~~~~') && codeAttachment.endsWith('~~~~~~~~~~')) {
+            codeAttachment = codeAttachment
+              .replace(/^~~~~~~~~~~/, '')
+              .replace(/~~~~~~~~~~$/, '')
+              .trim();
+          } else if (codeAttachment.startsWith('```') && codeAttachment.endsWith('```')) {
```
Jurredr, line 366: Just very small nitpick; 

If you're adding a comment `// our example` to indicate that this is how we do it, also add a comment here to just explain a bit of what you're doing here and why.

Jurredr (dismissed):

Jurredr (approved):

src/components/chat-item/chat-prompt-input.ts
```diff
@@ -355,11 +355,21 @@ export class ChatPromptInput {
...
+        if (codeAttachment.length > 0) {
+          // the way we mark code in our example mynah client
+          if (codeAttachment.startsWith('~~~~~~~~~~') && codeAttachment.endsWith('~~~~~~~~~~')) {
+            codeAttachment = codeAttachment
+              .replace(/^~~~~~~~~~~/, '')
+              .replace(/~~~~~~~~~~$/, '')
+              .trim();
+          } else if (codeAttachment.startsWith('```') && codeAttachment.endsWith('```')) {
+            // the way code is marked in VScode and JetBrains extensions
+            codeAttachment = codeAttachment
```
dogusata, line 368: This would break, the reason we added `~~~~~~~` here is to avoid this problem. What is coming from VSCode and JetBrains? We're not getting anything specific from VSCode or JetBrains other than pure text. Are we talking about the api response here?

dogusata (commented):
Please do not make a release with this one merged.

PR #205, opened 2024-12-16
Additional e2e tests for up down arrow navigation

francescoopiccoli:
## Problem
Add e2e test coverage for up/down arrow navigation, testing scenarios where the prompt is not sent, and when there is a code attachment inside the prompt.
We also rename some old tests for consistency with test naming. 

## Tests

## License


ui-tests/__test__/flows/navigate-prompts/navigate-back-to-current-prompt-with-code-attachment.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  let promptInput = await page.locator(`${getSelector(testIds.prompt.input)}`);
+  await promptInput.press('ArrowUp');
+  await waitForAnimationEnd(page);
+
+  promptInput = await page.locator(`${getSelector(testIds.prompt.input)}`);
+  await promptInput.press('ArrowDown');
+  await waitForAnimationEnd(page);
+
+  const codeAttachmentContent = (await page.locator(`${getSelector(testIds.prompt.attachment)}`).innerText()).trim();
```
francescoopiccoli, line 35: add to add `.trim()` because webpack test was failing otherwise as it adds a `\n` at the end, like `'This is an unsent code attachment\n'`

ui-tests/__test__/flows/navigate-prompts/stay-on-current-prompt.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const stayOnCurrentPrompt = async (page: Page, skipScreenshots?: boolean): Promise<void> => {
+  await closeTab(page, false, true);
+  await openNewTab(page, false, true);
+
+  // Write prompt without sending it
+  await page.locator(`${getSelector(testIds.prompt.input)}`).fill('This is the first unsent user prompt');
+  await waitForAnimationEnd(page);
+
+  let promptInput = await page.locator(`${getSelector(testIds.prompt.input)}`);
```
Jurredr, line 15: nit: define this selector for the prompt input as a `const`, instead of having duplicated code. Actually, I think you only have to define the locator once, and then this locator will keep updating. Try to see if that works.

ui-tests/__test__/flows/navigate-prompts/navigate-back-to-current-prompt-with-code-attachment.ts
Jurredr: Nice workaround. Add a comment above this line with this disclaimer as well for future code clarity.

ui-tests/__test__/flows/navigate-prompts/stay-on-current-prompt.ts
francescoopiccoli: Good idea, turns out the wrong screenshot height was related to defining the promptInput multiple times, as addressing this comment also fixed the screenshot height problem

Jurredr (dismissed):

ui-tests/__test__/flows/navigate-prompts/navigate-back-to-current-prompt-with-code-attachment.ts
```diff
@@ -0,0 +1,40 @@
...
+import { getSelector, waitForAnimationEnd } from '../../helpers';
+import testIds from '../../../../src/helper/test-ids';
+import { closeTab } from '../close-tab';
+import { openNewTab } from '../open-new-tab';
+
+export const navigateBackToCurrentPromptWithCodeAttachment = async (page: Page, skipScreenshots?: boolean): Promise<void> => {
+  await closeTab(page, false, true);
+  await openNewTab(page, false, true);
+
+  await page.locator(`${getSelector(testIds.prompt.input)}`).fill('This is the first user prompt');
```
natdurlik, line 11: nit: redundant backticks

francescoopiccoli:
Tests are failing because of a difference in screenshot height, if I run **only** the up/down arrow navigation e2e tests (as I did when pushing the latest test screenshot changes), the screenshot are shorter, if I run all e2e tests, the screenshots are higher for some reason. @Jurredr if it is fine to you I would keep the higher screenshot, as the content of the screenshot is correct, the only issue causing the fail is the screenshot height. I also find the longer screenshots to provide more context as they show all the chat items/prompt previously sent.

ui-tests/__test__/flows/navigate-prompts/stay-on-current-prompt.ts
francescoopiccoli: Actually I found out that that the two issues are not correlated, the fact that I got shorter screenshots is due to the fact that I reran only the up/down arrow navigation e2e tests.
(https://github.com/aws/mynah-ui/pull/205#issuecomment-2547957582)

natdurlik (approved):

ui-tests/__test__/flows/navigate-prompts/navigate-back-to-current-prompt-with-code-attachment.ts
Jurredr: Good nit, we have them quite a bit throughout the E2E tests. Will try to clean them up in my upcoming PR.

Jurredr (approved):

PR #208, opened 2024-12-16
feat: up/down arrow navigation now only triggers when cursor is at the start or end of multiline inputs

Jurredr:
## Problem
If there is multiline text in prompt field, it is not possible to navigate to upper lines by using app arrow, it triggers historical navigation.

## Solution
When there is text in the prompt field which is more than 1 line, until user approaches to the first line, keep the default behavior of the arrow keys.

## Tests

## License

### DIFF
--- a/src/components/chat-item/chat-prompt-input.ts
+++ b/src/components/chat-item/chat-prompt-input.ts
@@ -324,6 +324,26 @@ export class ChatPromptInput {
           this.quickPickOpen = true;
         }
       } else if (navigationalKeys.includes(e.key)) {
+        const inputText = this.promptTextInput.getTextInputValue();
+        const cursorPosition = this.promptTextInput.getCursorPos();
+
+        const textBeforeCursor = inputText.substring(0, cursorPosition);
+        const textAfterCursor = inputText.substring(cursorPosition);
+
+        // Check if cursor is on the first line
+        const isFirstLine = !textBeforeCursor.includes('\n');
+
+        // Check if cursor is on the last line
+        const isLastLine = !textAfterCursor.includes('\n');
+
+        // Only allow up/down arrow navigation when user reaches the end of multiline input
+        if (e.key === KeyMap.ARROW_UP && !isFirstLine) {
+          return;
+        }
+        if (e.key === KeyMap.ARROW_DOWN && !isLastLine) {
+          return;
+        }
+
         if (this.userPromptHistoryIndex === -1 || this.userPromptHistoryIndex === this.userPromptHistory.length) {
           this.lastUnsentUserPrompt = {
             inputText: this.promptTextInput.getTextInputValue(),


francescoopiccoli:
Personal preference/nit : From a UX perspective, I preferred having the arrows solely for prompt navigation, making it much faster to navigate through prompts. If the user wants to navigate (and modify) the input, it can already do so by moving the cursor with the mouse or using the left/right arrows. Have we validated this decision with UX?

src/components/chat-item/chat-prompt-input.ts
```diff
@@ -324,6 +324,26 @@ export class ChatPromptInput {
...
+        const cursorPosition = this.promptTextInput.getCursorPos();
+
+        const textBeforeCursor = inputText.substring(0, cursorPosition);
+        const textAfterCursor = inputText.substring(cursorPosition);
+
+        // Check if cursor is on the first line
+        const isFirstLine = !textBeforeCursor.includes('\n');
+
+        // Check if cursor is on the last line
+        const isLastLine = !textAfterCursor.includes('\n');
```
francescoopiccoli, line 337: just to double check, is it safe to deduce first/last line by checking for `\n`? i am thinking of some edge cases with input copy pasted starting/ending with empty lines

viktorsaws:
> Personal preference/nit : From a UX perspective, I preferred having the arrows solely for prompt navigation, making it much faster to navigate through prompts. If the user wants to navigate (and modify) the input, it can already do so by moving the cursor with the mouse or using the left/right arrows. Have we validated this decision with UX?

Personal IMO: using mouse breaks "typing" flow and maybe more jarring experience in non-terminal visual UI mode. Given this is multi-line input field, it is not as intuitive UI, as I expect to move cursor when I type. Maybe having special short cut to history jump would be more intuitive

src/components/chat-item/chat-prompt-input.ts
francescoopiccoli: In fact, I tried with starting/ending the input with some empty lines (shift + return), its quite a rare edge case i guess though

Jurredr:
> > Personal preference/nit : From a UX perspective, I preferred having the arrows solely for prompt navigation, making it much faster to navigate through prompts. If the user wants to navigate (and modify) the input, it can already do so by moving the cursor with the mouse or using the left/right arrows. Have we validated this decision with UX?
> 
> Personal IMO: using mouse breaks "typing" flow and maybe more jarring experience in non-terminal visual UI mode. Given this is multi-line input field, it is not as intuitive UI, as I expect to move cursor when I type. Maybe having special short cut to history jump would be more intuitive

I've been thinking about this as well. A key combination for history jumps might be better. Will get back to this when I'm back in office on Thursday.

bergjaak:
> Personal preference/nit : From a UX perspective, I preferred having the arrows solely for prompt navigation, making it much faster to navigate through prompts. If the user wants to navigate (and modify) the input, it can already do so by moving the cursor with the mouse or using the left/right arrows. Have we validated this decision with UX?

For users whose flow does not involve using a mouse, it's annoying to have to use a mouse to edit the prompt. Also, how often does a user want to scrollback through their prompt history? If they need to go far, they can just scroll up in the chat... and I wonder how rare that is relative to needing to edit the prompt

src/components/chat-item/chat-prompt-input.ts
francescoopiccoli: I did some more testing, and I think it is not a safe deduction to check for `'\n'` to deduce first/last line. To do this we might need to check the height of the input field or count the n of lines somehow

dogusata: Agree with Francesco.

PR #209, opened 2024-12-19
Add E2E tests for voting and feedback form

Jurredr:
## Problem
There were no end to end tests for voting and feedback forms yet.

## Solution
Added 6 new flows for testing this E2E.

## Extra
Cleaned up a bunch of unnecessary use of backticks (`) and `await` keywords throughout all E2E tests.

## Tests (NA)
~- [ ] I have tested this change on VSCode~
~- [ ] I have tested this change on JetBrains~

## License


ui-tests/__test__/flows/feedback-form/render-upvote-result.ts
```diff
@@ -0,0 +1,31 @@
...
+        body: 'This message is votable.',
+        canBeVoted: true,
+      });
+    }
+  });
+  await waitForAnimationEnd(page);
+
+  const thumbsUp = page.locator(getSelector(testIds.chatItem.vote.upvoteLabel));
+  expect(thumbsUp).toBeDefined();
+  await thumbsUp.click();
```
francescoopiccoli, line 25: nit: should we check (with expect(..)) also for what happens after the click? namely the UI change with thanks message?

ui-tests/__test__/flows/feedback-form/render-vote-buttons.ts
```diff
@@ -0,0 +1,30 @@
...
+export const renderVoteButtons = async (page: Page, skipScreenshots?: boolean): Promise<void> => {
+  await closeTab(page, false, true);
+  await openNewTab(page, false, true);
+
+  await page.evaluate((body) => {
+    const selectedTabId = window.mynahUI.getSelectedTabId();
+    if (selectedTabId != null) {
+      window.mynahUI.updateStore(selectedTabId, {
+        chatItems: [],
+      });
```
francescoopiccoli, line 15: nit: this check looks redundant as we open a new tab, but ok with keeping it. In general I wonder whether it's better for the other new tests, to close tab and open tab at the beginning of the test, compared to resetting the chatItems in the store, not sure if there is a difference, I dont have strong opinions

ui-tests/__test__/flows/feedback-form/submit-feedback-form.ts
```diff
@@ -0,0 +1,40 @@
...
+  const reportButton = page.locator(getSelector(testIds.chatItem.vote.reportButton));
+  expect(reportButton).toBeDefined();
+  await reportButton.click();
+  await waitForAnimationEnd(page);
+
+  const submitButton = page.locator(getSelector(testIds.feedbackForm.submitButton));
+  expect(submitButton).toBeDefined();
+  await submitButton.click();
+  await waitForAnimationEnd(page);
+
```
francescoopiccoli, line 36: nit: should we also check for the feedback form to be dismissed after this action?

ui-tests/__test__/flows/feedback-form/cancel-feedback-form.ts
francescoopiccoli, line 1: this test depends on the previous one (render feedback form), is that ok or should we make it 'independent' as we do for submit-feedback-form?

ui-tests/__test__/flows/feedback-form/cancel-feedback-form.ts
```diff
@@ -0,0 +1,14 @@
+import { Page } from 'playwright/test';
+import { getSelector, waitForAnimationEnd } from '../../helpers';
+import testIds from '../../../../src/helper/test-ids';
+
+export const cancelFeedbackForm = async (page: Page, skipScreenshots?: boolean): Promise<void> => {
+  const cancelButton = page.locator(getSelector(testIds.feedbackForm.cancelButton));
+  expect(cancelButton).toBeDefined();
+  await cancelButton.click();
+  await waitForAnimationEnd(page);
+
```
francescoopiccoli, line 10: nit: should we also check for the feedback form to be dismissed after this action?

francescoopiccoli (approved):
Overall great work! Just a few comments, mostly from my understanding, I approve but make sure to remove the changes in chat-prompt-input

ui-tests/__test__/flows/feedback-form/render-upvote-result.ts
Jurredr: We check this with the screenshot already, so no point in double testing this really.

ui-tests/__test__/flows/feedback-form/cancel-feedback-form.ts
Jurredr: We do this with the screenshot comparison.

ui-tests/__test__/flows/feedback-form/submit-feedback-form.ts
Jurredr: We check this as well with the screenshot itself.

ui-tests/__test__/flows/feedback-form/render-vote-buttons.ts
Jurredr: We've noticed that closing and opening tabs for each flow causes issues with flakiness and is unstable, so I try to only do it on the first one of a test group, and then only doing it if a full reset is actually necessary and can't be done easily through the store.

Jurredr: But you're definitely right on that initial store update not being necessary, removed it.

ui-tests/__test__/flows/feedback-form/cancel-feedback-form.ts
Jurredr: Yeah not 100% sure here either, I thought it'd be fine as the whole `describe` group makes up the E2E test for this feature, and we're not randomly going to swap out one flow or test it in another place. But not sure what the 'correct' way or standard way of doing it is here. Will ask @dogusata when he's back, for now let's keep it like this.

francescoopiccoli (approved):

PR #211, opened 2024-12-20
feat: new syntax highlighter

Jurredr:
## Problem
The current syntax highlighter lacks proper tokenization control, thus leading to unattractive syntax highlighting.

## Solution
Swap out `prism.js` by `highlight.js`. The library `shiki.js` has also been considered, but has not been used given its strictly asynchronous nature.  

### Old
Screenshot 2024-12-20 at 18 09 11

### New
Screenshot 2024-12-20 at 18 09 11

## Important note
A proper color mapping still needs to be added, possibly a dynamic one consuming variables from VSCode.

## Tests

## License


src/styles/_variables.scss
```diff
@@ -30,39 +30,56 @@
     --mynah-color-toggle-reverse: rgba(0, 0, 0, 0.5);
 
     --mynah-color-syntax-bg: var(--vscode-terminal-dropBackground);
-    --mynah-color-syntax-variable: var(
```
dogusata, line 33: These should remain as is, the theme you'll use for the highlight.js should use those css custom properties.
We're not gonna get rid of these custom properties. The reason we have those is to not make any change on the usage of the MynahUI even if we change the library we use, like in this case from prism.js to highlight.js.

dogusata: And by default, we're using VSCode theme colors, however you can add all new color properties as fallbacks:
For example:

```css
--mynah-color-syntax-variable: var(--vscode-gitDecoration-modifiedResourceForeground,
        var(--vscode-debugTokenExpression-name, #986801))
    )
```

dogusata (changes requested):

dogusata:
2 more important things:

- Please triple check if the reference tracker highlights are still working as expected (which are already part of the e2e tests, however checking manually would also be beneficial here)
- Please update the third party licenses file, since this is an open source library, it has to have those information on the root. Remove prismjs and add highlightjs (and any additional third part library you put to make highlightjs work as expected)

package.json
```diff
@@ -36,17 +36,17 @@
     },
     "dependencies": {
         "escape-html": "^1.0.3",
+        "highlight.js": "^11.11.0",
         "just-clone": "^6.2.0",
-        "marked": "^14.1.0",
-        "prismjs": "1.29.0",
+        "marked": "^12.0.2",
```
dogusata, line 41: Any specific reason we downgrade the version of marked?

src/helper/merge-html-plugin.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    function selectStream (): Event[] {
+      if ((original.length === 0) || (highlighted.length === 0)) {
+        return (original.length > 0) ? original : highlighted;
+      }
+      if (original[0].offset !== highlighted[0].offset) {
+        return (original[0].offset < highlighted[0].offset) ? original : highlighted;
+      }
+
+      /*
```
dogusata, line 95: We don't need code comments I believe.

src/main.ts
```diff
@@ -268,10 +268,9 @@ export class MynahUI {
     // Apply global fix for marked listitem content is not getting parsed.
     marked.use({
       renderer: {
-        listitem: (item: Tokens.ListItem) => `
+        listitem: (text: string, task: boolean, checked: boolean) => `
```
dogusata, line 271: This would be changed with the proper version of marked. (14.*)

src/styles/components/syntax/_syntax-theme.scss
```diff
@@ -0,0 +1,88 @@
+.diff {
```
dogusata, line 1: Why not using the _syntax-highlighter.scss? It is already targets to that. Theme doesn't provide any beneficial thing here. Just using the main scss would be fine.

dogusata (changes requested):

src/helper/merge-html-plugin.ts
Jurredr: This plugin is taken from https://github.com/highlightjs/highlight.js/issues/2889. I think we should just keep the source as is, I just made it TypeScript compatible. I credited the source at the top.

package.json
Jurredr: Seemed like it was a requirement for highlight.js. I'll double check if we can get it to stay at 14.1.0.

src/styles/components/syntax/_syntax-theme.scss
Jurredr: Good catch, forgot to clean this up actually. It was a separate file before as I just imported an existing theme file from highlight.js to experiment with.

package.json
Jurredr: Think I managed to bump it up again without issues. Let's see.

src/helper/merge-html-plugin.ts
dogusata: If we're cloning it into our repo instead of using it directly from the node_modules as a dependency, we don't need these comments. Please avoid it, instead just put a reference in the THIRD PARTY file.

dogusata (approved):
Added just a small nit related with the 3rd party clone usage and comments. Please remove the comments in your next pr, if we're not using it directly from node_modules for some reason, the code should follow our guidelines not anyone elses.

PR #214, opened 2024-12-24
Added cursor position check before running prompt history trigger

dogusata:
## Problem
- When using the up and down arrow keys to grab previous prompts like in a terminal, it is not possible to navigate through the text anymore. It should only trigger the historical previous/next prompt actions when on the first and last lines
- Cursor jumps to text end after up and down arrow actions even if there is no previous prompt available.

## Solution
- Added current cursor line check before triggering historical prompt calls.
- Fixed unnecessary cleanup which causes a cursor jump to the end.

## Tests

## License

### DIFF
--- a/src/components/chat-item/chat-prompt-input.ts
+++ b/src/components/chat-item/chat-prompt-input.ts
@@ -331,15 +331,16 @@ export class ChatPromptInput {
           };
         }
 
-        this.clearTextArea();
-
         if (this.userPromptHistoryIndex === -1) {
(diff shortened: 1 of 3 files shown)


ege0zcan (approved):
I would suggest adding some test cases to the ui tests to check for some cases:
- Going up and down in a multiline prompt
- Going up when the prompt input area is empty
- Test case where the cursor is initially in the first line and then window size is shrunk down and the cursor moves to the next line

PR #215, opened 2024-12-24
Add additional backslash to escapedPrompt

rahmaniaam:
## Problem
Chat prompts with a sequence of multiple backslashes will result in backslashes to be gone in the prompt card

Input example:
```
Make a regex that get '\\(' occurrences on string

Make a regex that get \\( occurrences on string

Make a regex that get\\( occurrences on string

Make a regex that get \\(occurrences on string

Make a regex that get '\\\(' occurrences on string

Make a regex that get '\\' occurrences on string

Make a regex that get \\ occurrences on string

Make a regex that get \' occurrences on string

Make a regex that get '\ occurrences on string
```

Result:
Screenshot 2024-12-24 at 15 37 00

## Solution
Add regex logic in the escaped prompt to add an additional backslash on sequences of backslashes

Result:
Screenshot 2024-12-24 at 15 35 24

## Tests

## License

### DIFF
--- a/src/components/chat-item/chat-prompt-input.ts
+++ b/src/components/chat-item/chat-prompt-input.ts
@@ -679,7 +679,7 @@ export class ChatPromptInput {
           context.push(match);
         }
         return `**${match}**`;
-      }));
+      })).replace(/\\/g, '\\\\');
 
       const promptData: {tabId: string; prompt: ChatPrompt} = {
         tabId: this.props.tabId,
--- a/src/components/chat-item/chat-prompt-input.ts
+++ b/src/components/chat-item/chat-prompt-input.ts
@@ -679,7 +679,7 @@ export class ChatPromptInput {
           context.push(match);
         }
         return `**${match}**`;
-      })).replace(/\\/g, '\\\\');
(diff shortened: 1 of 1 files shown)


Jurredr:
Nice fix with the regex, seems to work as expected. Just wondering; do we know what exactly is the root cause of this? Is there not something we can do to prevent it in the first place, rather than patching it up with a replacement?

rahmaniaam:
> Just wondering; do we know what exactly is the root cause of this? Is there not something we can do to prevent it in the first place, rather than patching it up with a replacement?

From what I can tell, it doesn't seem that the `escapeHTML` function successfully interpret the backslashes resulting in the card to render incorrectly, we may need to use a different library

rahmaniaam:
This needs more work, playing around with escaping backticks is still messing it up:

Input
```
Make a regex that get `\``\\(``` occurrences on string
```
or
```
Make a regex that get ```\\(``\` occurrences on string
```
Output
```
Make a regex that get ```\(``` occurrences on string
```

dogusata:
The current behavior works as expected for a markdown input field. It is also the exact same behavior for the Github or any other markdown supported entry field in general.

PR #218, opened 2025-01-08
feat: improved random UID generation through crypto module

Jurredr:
## Problem
Even though cryptographic safety is not really relevant to anything we're doing with IDs, it's good practice to make use of Node's built-in crypto package anyways, over implementing our own less optimized solution. See https://github.com/aws/mynah-ui/security/code-scanning/8.

## Solution
Use crypto's random UUID function.

## Tests

## License

### DIFF
--- a/src/helper/guid.ts
+++ b/src/helper/guid.ts
@@ -4,7 +4,5 @@
  */
 
 export const generateUID = (): string => {
-  const firstPart: number = (Math.random() * 46656) | 0;
-  const secondPart: number = (Math.random() * 46656) | 0;
-  return `000${firstPart.toString(36)}`.slice(-3) + `000${secondPart.toString(36)}`.slice(-3);
+  return crypto.randomUUID();
 };


dogusata:
Does it require a third party dependency?

Also, we're not using the UUIDs for specific identification and not storing them in the client. I'm not so sure about changing the current ID structure.

Jurredr:
> Does it require a third party dependency?
> 
> Also, we're not using the UUIDs for specific identification and not storing them in the client. I'm not so sure about changing the current ID structure.

I think it's better for the future to use the standard, instead of our own implementation. I don't see a need to reinvent the wheel. `crypto` is part of NodeJS, it's not like a separate new third party dependency, and it's fully stable and never changed (https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID).

Jurredr:
Closing this. Unnecessary hassle, I'm seeing that it has some ifs and buts like only working properly with https possibly. Also it's already failing on the UI testing package.

PR #222, opened 2025-01-10
Dogusata/fix cursor position reset after history item set

dogusata:
## Problem
- When you use the cursor to navigate between the older prompts, after the prompt is set, the cursor position resets itself to end of the text after 100-500ms.
- When there is a code attachment in the historical prompts, it was not possible to navigate anymore directly with up/down arrow. Also, the code attachment remains on screen.

## Solution
- Setting the prompt field text through global store was not necessary, since it is an async process. Instead we have the `promptTextInput` component already in hand, which has the `updateText` method.
- Setting the code attachment through global store was not necessary since it is an async process . Instead we have the `promptAttachment` component already in hand, which has the `updateAttachment` and `clear` methods. **However,** it might seem like making it through the data store can also do the checks for max length with prompt text and the attachment. **BUT** since users cannot send prompts without this check it means that the historical prompts already being checked before they are sent. **Again, only updating the component is enough**.
- Else state was missing for the case if there is no attachment on the previous or last not sent prompts. If there is no attachment, it should clear the code attachment field.

## Tests

## License

### DIFF
--- a/src/components/chat-item/chat-prompt-input.ts
+++ b/src/components/chat-item/chat-prompt-input.ts
@@ -346,14 +346,10 @@ export class ChatPromptInput {
 
         let codeAttachment = '';
         if (this.userPromptHistoryIndex === this.userPromptHistory.length) {
-          MynahUITabsStore.getInstance().getTabDataStore(this.props.tabId).updateStore({
-            promptInputText: this.lastUnsentUserPrompt.inputText ?? '',
-          });
+          this.promptTextInput.updateTextInputValue(this.lastUnsentUserPrompt.inputText ?? '');
           codeAttachment = this.lastUnsentUserPrompt.codeAttachment ?? '';
         } else {
-          MynahUITabsStore.getInstance().getTabDataStore(this.props.tabId).updateStore({
-            promptInputText: this.userPromptHistory[this.userPromptHistoryIndex].inputText,
-          });
+          this.promptTextInput.updateTextInputValue(this.userPromptHistory[this.userPromptHistoryIndex].inputText);
           codeAttachment = this.userPromptHistory[this.userPromptHistoryIndex].codeAttachment ?? '';
         }
         codeAttachment = codeAttachment.trim();
@@ -371,7 +367,9 @@ export class ChatPromptInput {
               .replace(/```$/, '')
               .trim();
           }
-          this.addAttachment(codeAttachment, 'code');
+          this.promptAttachment.updateAttachment(codeAttachment, 'code');
+        } else {
+          this.promptAttachment.clear();
         }
       }
     } else {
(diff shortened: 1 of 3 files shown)


Jurredr (approved):

src/components/chat-item/chat-prompt-input.ts
```diff
@@ -346,14 +346,10 @@ export class ChatPromptInput {
 
         let codeAttachment = '';
         if (this.userPromptHistoryIndex === this.userPromptHistory.length) {
-          MynahUITabsStore.getInstance().getTabDataStore(this.props.tabId).updateStore({
-            promptInputText: this.lastUnsentUserPrompt.inputText ?? '',
-          });
+          this.promptTextInput.updateTextInputValue(this.lastUnsentUserPrompt.inputText ?? '');
```
ege0zcan, line 349: just to be sure, there is no chance that the unsent text can go over the word limit right?

ege0zcan (approved):

src/components/chat-item/chat-prompt-input.ts
```diff
@@ -371,7 +367,9 @@ export class ChatPromptInput {
               .replace(/```$/, '')
               .trim();
           }
-          this.addAttachment(codeAttachment, 'code');
+          this.promptAttachment.updateAttachment(codeAttachment, 'code');
+        } else {
+          this.promptAttachment.clear();
```
ege0zcan, line 372: Maybe this was intentional, for example imagine the scenario when I write a message asking about my code and I get a weak response, I then decide to send the same message with the context attached so I attach the code to the prompt and click up to get my previous prompt. I am guessing this is not a great user experience but wanted to explicitly discuss the intent here

src/components/chat-item/chat-prompt-input.ts
dogusata: Yes, while typing they're being checked. Even if it is not sent yet, it still cannot exceed the max word count.

src/components/chat-item/chat-prompt-input.ts
francescoopiccoli: this behavior was alredy there before the last fix to trigger navigation only on first/last line (see: https://github.com/aws/mynah-ui/pull/214/files#diff-bd4c5481ffce0a6058196b3d09baac811eb95bae5f07b53a61ff3cdf3658323fL334). I prefer not dragging the attached code across previous prompts, there are more scenarios where this behavior is not intended by the user, than the opposite imo. it also introduces weird behavior when a previous prompt actually had a code attachment. for the use case you described, the user can just navigate back to previous prompt and then attach the code!

ege0zcan: Nice, thanks for the clarification

PR #224, opened 2025-01-16
221 pressing tab on context selector duplicates the subsequent text if it exists

dogusata:
## Problem
- When tab is changed or in VSCode Q panel gets closed and opened back, scroll position randomly changes
- When context selector panel is opened (hitting @) if the next character on the cursor position is not a space or end of line filtering doesn't work and if user hits enter | tab | space it duplicates the next word available.

#### Scroll position misbehaviors:
When VSCode hides an extension panel, it shrinks the width and height of it to 0. And since the container size also shrinks down, it was setting the scroll position to an unexpected one. We added a `ROOT_RESIZE` global event and the handler for it in the `ChatWrapper` to understand if the root element size goes under 10px width or 10px height when that tab is active. When root elm gets resized back (more than 10px width and height but only once until the next lower than 10px), we're setting the scroll position back to the latest recorded before the hide precedure.
The same applies for the tab switches. When a tab is deselected, it looses height and width which causes scroll position to go to a random value. We added a new subscription on tab store model called `beforeTabChange` and in the handler we're recording the last scroll position. When the tab get selected back we're resetting the scroll position to the latest recorded.

#### Context selector misbehaviors:
As user types after opening the context selector, the filtering was checking the written text from the cursor start to next space or line end character not what the actual cursor goes. Now it checks depending on the cursor position. So as user types filtering matches what user types instead of trying to match the whole word starting from the init cursor index. When user hits tab, space or enter, it only adds the matching context item. If there is no context item in the filtered view, it just leaves it as is.

## Tests

## License


src/components/chat-item/chat-wrapper.ts
```diff
@@ -74,6 +77,32 @@ export class ChatWrapper {
         this.allRenderedChatItems = {};
       }
     });
+
+    MynahUITabsStore.getInstance().addListener('selectedTabChange', (selectedTabId) => {
+      if (this.props.tabId === selectedTabId) {
+        setTimeout(() => {
+          this.chatItemsContainer.scrollTop = this.scrollPos;
```
dogusata, line 84: The delay is mandatory to let the renderer finish the scroll positioning first. 1ms would also be fine, but 1ms can be skipped for some browser in some conditions.

src/components/chat-item/chat-wrapper.ts
```diff
@@ -74,6 +77,32 @@ export class ChatWrapper {
...
+      }
+    });
+
+    MynahUIGlobalEvents.getInstance().addListener(MynahEventNames.ROOT_RESIZE, (clientRect: ReturnType<HTMLElement['getBoundingClientRect']>) => {
+      if ((clientRect.height < 10 || clientRect.width < 10) && this.props.tabId === MynahUITabsStore.getInstance().getSelectedTabId()) {
+        this.restoreScrollAfterResize = true;
+        this.scrollPos = this.chatItemsContainer.scrollTop;
+      } else if (!this.restoreScrollAfterResize) {
+        this.restoreScrollAfterResize = false;
+        setTimeout(() => {
```
dogusata, line 100: The delay is mandatory to let the renderer finish the scroll positioning first. 1ms would also be fine, but 1ms can be skipped for some browser in some conditions.

src/helper/tabs-store.ts
```diff
@@ -101,8 +104,7 @@ export class MynahUITabsStore {
   public updateTab = (tabId: string, tabData?: Partial<MynahUITabStoreTab>, skipSubscribers?: boolean): void => {
     if (this.tabsStore[tabId] !== undefined) {
       if (tabData?.isSelected === true && this.getSelectedTabId() !== tabId) {
-        this.deselectAllTabs();
-        this.informSubscribers('selectedTabChange', tabId);
+        this.selectTab(tabId);
```
dogusata, line 107: We already had a function for that, unnecessary to duplicate internally.

src/components/chat-item/chat-wrapper.ts
```diff
@@ -74,6 +77,32 @@ export class ChatWrapper {
...
+
+    MynahUITabsStore.getInstance().addListener('selectedTabChange', (selectedTabId) => {
+      if (this.props.tabId === selectedTabId) {
+        setTimeout(() => {
+          this.chatItemsContainer.scrollTop = this.scrollPos;
+        }, 10);
+      }
+    });
+    MynahUITabsStore.getInstance().addListener('beforeTabChange', (selectedTabId) => {
+      if (this.props.tabId !== selectedTabId) {
```
dogusata, line 89: Just before the tab change, we're updating `this.scrollPos` which holds the last scroll position.

src/components/chat-item/chat-wrapper.ts
```diff
@@ -74,6 +77,32 @@ export class ChatWrapper {
...
+        this.scrollPos = this.chatItemsContainer.scrollTop;
+      }
+    });
+
+    MynahUIGlobalEvents.getInstance().addListener(MynahEventNames.ROOT_RESIZE, (clientRect: ReturnType<HTMLElement['getBoundingClientRect']>) => {
+      if ((clientRect.height < 10 || clientRect.width < 10) && this.props.tabId === MynahUITabsStore.getInstance().getSelectedTabId()) {
+        this.restoreScrollAfterResize = true;
+        this.scrollPos = this.chatItemsContainer.scrollTop;
```
dogusata, line 97: When the window is resized to a viewport smaller than 10px by 10px, we're updating `this.scrollPos` which holds the last scroll position. **We're able to do that because the `ResizeObserver` trigger runs before the scroll reset.**

src/components/chat-item/chat-wrapper.ts
```diff
@@ -74,6 +77,32 @@ export class ChatWrapper {
...
+      } else if (!this.restoreScrollAfterResize) {
+        this.restoreScrollAfterResize = false;
```
Jurredr, line 99: This never has an effect right? You set it to false only if it was already false in the first place

Jurredr (dismissed):

src/components/chat-item/chat-wrapper.ts
dogusata: Yes it should be if(this.restoreScrollAfterResize)

Jurredr (dismissed):

Jurredr (approved):

PR #226, opened 2025-01-23
added link rendering handler to markedjs

dogusata:
## Problem
When links arrive inside markdown content as pure text, by default they're being converted as anchor elements. However, it is not safe for Amazon Q Dev, only the `[]()` formatted links should appear as links and the rest should remain as text.

## Solution
Added link renderer handler for the markedjs.
Screenshot 2025-01-23 at 20 13 48

PS:
- e2e tests needs to be added (send body with pure text link and `[]()` formatted link)

## Tests

## License

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -39,6 +39,7 @@ import { generateUID } from './helper/guid';
 import { NoTabs } from './components/no-tabs';
 import { copyToClipboard } from './helper/chat-item';
 import { Spinner } from './components/spinner/spinner';
+import unescapeHTML from 'unescape-html';
 
 export { generateUID } from './helper/guid';
 export {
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -272,7 +273,31 @@ export class MynahUI {
...
+              .replace(/[^\w:]/g, '')
+              .toLowerCase();
+          } catch (e) {
+            return '';
+          }
+          if (prot.indexOf('javascript:') === 0 || prot.indexOf('vbscript:') === 0 || prot.indexOf('data:') === 0) {
+            return '';
+          }
+
+          if (item.href === item.text && item.title == null) {
```
justinmk3, line 291: if `item.title` is empty string or undefined, that probably isn't useful to set in `title="..."`, right?
```suggestion
          if (item.href === item.text && !!item.title) {
```

Jurredr (approved):

PR #227, opened 2025-01-31
added description field to feedback form

dogusata:
## Problem
- Feedback form is missing description field to inform the end user or let them take some actions like link clicks

## Solution
- Added description field which can be set through the `config`. It accepts markdown format content including links. If there is a link in it, it will be passed to the consumer via `onFormLinkClick` constructor event handler property.
- Updated the documentation
- Updated the tests for feedback form
- Updated example app

image

## Tests

## License


docs/PROPERTIES.md
```diff
@@ -745,6 +749,30 @@ onLinkClick?: (
 ```
 ---
 
+### `onFormLinkClick`
+
+This event will be fired when user clicks a link inside the description field on feedback or custom forms. It will pass `link`, for the clicked link and the `mouseEvent` and the to check userIntent the `eventId`.
```
ege0zcan, line 754: nit typo: ``and the to check userIntent the `eventId`.``

dogusata: noted for next PR

PR #230, opened 2025-02-05
Export chats to markdown or HTML

Jurredr:
## Problem
Users would like to export their chats so that they can revisit them later.

## Solution
- Added a serialization function to serialize a tab to an HTML or markdown string
- For markdown, it will list all card body strings, separated by an empty line and `---`
- For HTML, it will include the MynahUI stylesheets to replicate the chat window, with all buttons / interactions removed from the cards:
Screenshot 2025-02-05 at 14 10 50

The example app includes an example of calling this serialization and using it to download a file.

## Tests (NA)
~- [ ] I have tested this change on VSCode~
~- [ ] I have tested this change on JetBrains~

## License


src/main.ts
```diff
@@ -727,6 +728,18 @@ ${(item.task ? marked.parseInline : marked.parse)(item.text, { breaks: false })
     }
   };
 
+  /**
+   * Serialize all (non-empty) chat messages in a tab into a string
+   * @param tabId Corresponding tab ID.
+   * @param fileType Whether to serialize to markdown or HTML format
+   */
+  public serializeChat = (tabId: string, fileType: 'markdown' | 'html'): string => {
```
dogusata, line 736: nit: just `type` or `contentType` would be better, since we're sending content instead of an actual file.

dogusata (approved):
Brilliant!

src/main.ts
Jurredr: Totally right, thanks, will fix

dogusata:
Maybe also removing the code block actions and then releasing it would be a better idea.

PR #232, opened 2025-02-11
Added conversation persistence methods on client side

Jurredr:
## Problem
There was no (documented) way to save and load conversations / tabs on the client side within MynahUI.

## Solution
Added config options to persist the chat interactions (across all tabs) locally with an overridable `save()` function. Also added and autoSave config option, which will save in 4 default cases (see documentation).

## Tests

## License

### DIFF
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -323,6 +323,21 @@ default: `false`
 
 ---
 
+# `autoSave`
+Whether to run the `MynahUI.save()` function by default on a couple of predefined events:
+- Whenever a new tab is created
+- Whenever a tab is closed
+- Whenever a tab is selected
+- Whenever any chatItem is added to a chat
+
+This will ensure a very basic saving behavior, but is not optimized nor specialized for custom saving implementations. You can extend the default saving behavior by adding `mynahUI.save()` to specified event handlers in the `MynahUI` constructor.
+
+**Note: if `persistConversations` is set to false, `autoSave` naturally also does not have any effect.**
+
+default: `false`
+
+---
+
 ## `codeInsertToCursorEnabled` and `codeCopyToClipboardEnabled` (default: true)
 These two parameters allow you to make copy and insert buttons disabled system wide. If you want to disable it specifically for a message you can do it through `ChatItem` object. Please see [DATAMODEL Documentation](./DATAMODEL.md#codeinserttocursorenabled-and-codecopytoclipboardenabled-default-true).
 
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -320,6 +320,9 @@ default: `4096`
 # `persistConversations`
 Whether data of tabs should be persisted so that it can be loaded later. If disabled, the `MynahUI.save()` function will have no effect. When enabled, the `MynahUI.save()` function will either save to localStorage by default under the `mynah-ui-storage` key, or use custom saving behavior defined with the `onSave()` callback.
 
+> [!NOTE]
+> Make sure to take into account how different IDE windows e.g. could conflict by saving to the same location, causing possibly unexpected saving behavior.
+
 default: `false`
 
 ---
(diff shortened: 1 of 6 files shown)


docs/CONFIG.md
```diff
@@ -47,6 +47,8 @@ interface ConfigModel {
     }>;
     tabBarButtons?: TabBarMainAction[]; // Tab bar buttons will be shown on the right of the tab
     maxUserInput: number; // max number of chars for the input field
+    persistConversations: boolean, // whether to save conversations, by default in localStorage
```
francescoopiccoli, line 50: is local storage available for browsers only or other clients as well (e.g. an ide)?

docs/CONFIG.md
```diff
@@ -315,6 +317,28 @@ default: `4096`
 
 ---
 
+# `persistConversations`
+Whether data of tabs should be persisted so that it can be loaded later. If disabled, the `MynahUI.save()` function will have no effect. When enabled, the `MynahUI.save()` function will either save to localStorage by default under the `mynah-ui-storage` key, or use custom saving behavior defined with the `onSave()` callback.
+
```
francescoopiccoli, line 322: do you think it could be useful to add  here or somewhere else) some sort of warning about the scenario/edge case where multiple mynah chats (not tabs) are open (e.g. in different browser tabs or different IDEs windows) and possible overwriting of the storage? ofc it depends on the storage location implemented by the consumers and its not up to us

docs/CONFIG.md
Jurredr: It does not persist properly in an iframe in a web client like VSCode, no. This is the default, which is at least accepted by all browser environments and works properly for testing purposes in a web browser. The expectation is that clients like IDEs will override with their own custom saving behavior.

francescoopiccoli: good to me, is it ok for us to have client-specific (i.e. browser-only (localStorage)) documentation in docs/config file or should we keep it general?

Jurredr: Well, that's why it's suggested that it's the default here. There will be clear communication with stakeholders on the exact steps they need to take.

dogusata:
General comment; even though i find the approach nice, I would prefer to move logical part of this implementation to consumer side by providing a proper documentation as you did now.

_**1st:**_ We shouldn't load the last saved state automatically. It should be called by the consumer after `onUIReady` event if they want. We can override a desired initial data instead. Provide `loadLocallySavedStore: ()=>boolean` (return `true` if load was successful) and do not automatically load it.

_**2nd:**_ I guess one of them is really not necessary: `autoSave` and `persistentStorage`. Instead, use only one with a name `persistStoreLocally`.

_**3rd:**_ I'm not sure about providing the `save` public method if we also provide `persistStoreLocally` in its current functionality. **But**, we actually still need it. It should accept a data argument in tabs' data model. Otherwise what is the reason to call the save manually? We know when the data is updated already, we can do it automatically. But if the consumer wants to override it with some kind of data actually stored on cloud and keep it aligned with local, they should be able to override what we stored locally. Also, please rename and re-type it to `saveStoreLocally: ()=>boolean` (and return `true` if saving was successful) which would be more understandable. Just `save` doesn't mean we're saving the current data store or sth more than that. It can be anything and also to anywhere.

**_4th_:** Rename the storage key name to a unique one time static value. And make it optional by providing a config item (`localStoreSavingKey: string;` or sth. similar)

**_5th_:** How the consumer can reset/clear the saved data? Assume that they decided to set the autoSave to false, the last saved state will remain forever until they override it. And overriding it with an empty data doesn't mean that it is not existent. There should be a `clearLocallySavedStore: ()=>boolean` (and return `true` if clear was successful) public method to allow it to be removed completely from storage. 

**_6th:_** How you're 100% sure that the `localStorage` will be there? `caniuse` or anything similar cannot provide you this information for custom webviews (JCEF, EdgeView etc.). We don't know what kind of browser functionality will be supported on all target platforms' webviews. We have now 4 target platforms and one of them can decide not to provide you localStorage in your isolated frame for an extension. Please do check the availability of the `localStorage` functionality first, and in case if it is not available: return **`false`** for `loadLocallySavedStore`, `saveStoreLocally` and `clearLocallySavedStore` methods. Since we don't a locally available store to save the data.

dogusata (changes requested):

francescoopiccoli (approved):

PR #234, opened 2025-02-16
Beta/falcon features

dogusata:
This set of features includes
- **Prompt input field**
  - Whole structure is changed with a content editable div, instead of previous textarea + sizer/visualizer span combination
  - Increased min-height to 3 lines if the screen height is equal to or more than 450px, otherwise keeps 1 line.
- **Context item and Quick action commands**
  - Added nested list functionality, now a command or context can have children which will have an indication of right arrow when user focuses/hovers to that item 
  - Improved filtering through nested items and highlight matching
  - Added `actions` option for the group, which is being placed on top of the group title
  - Selected context items are now being sent as an array of context item objects with the `onChatPrompt` event
  - Added a new event called `onContextSelected` which uses the return of the handler as a condition to trigger the default behavior or not. Added to make context/command items as an actionable item like a button.
  - Added icon (if available) to the selected context items
  - Added tooltip to the selected context items
  - Added autoscroll to element while navigating with up and down arrows in the list if the item is not in the visible scroll area.
- **Chat item and chat card**
  - Added `header` option to chat item, works similar with footer, however to avoid re-rendering, it requires a `undefined` as a value at least once while updating the chat item.
- **File list/tree**
  - Added `collapsed` option to collapse all folders 
  - Added option to hide file count
  - Added option to make the file list flat instead of a tree (which hides folders except the root)
  - Added vertical scroll instead of truncation.
  - Added folder icon option. If `null` is given no folder icons will be showed.
- **Forms** _(both in ChatItem and custom form field)_
  - Added validation pattern check for textual items. They can be multiple with a regex pattern array and can have individual error messages or one message for any of them. They can also be compared with `OR` or `AND`.
  - Added modifier key (cmd or ctrl) + enter in textual items to trigger a new event `onFormModifierEnterPress` to be able to submit the form.

## Tests

## License


src/styles/components/chat/_chat-prompt-wrapper.scss
```diff
@@ -140,15 +133,25 @@
...
                             position: absolute;
-                            left: -1px;
-                            right: -1px;
+                            left: 0;
+                            right: 0;
+                            top: 0;
+                            bottom: 0;
                             width: auto;
-                            height: 100%;
+                            height: calc(100%);
```
Jurredr, line 152: Unnecessary calc

Jurredr (approved):

PR #235, opened 2025-02-17
Added header field to chatitem

Jurredr:
## Problem
The ChatItem should have a header attribute identical to footer but appears on top of the chat card instead of the footer.

## Solution
Added a header prop, similar to the existing footer.

## Tests

## License


src/components/chat-item/chat-item-card.ts
```diff
@@ -190,6 +200,39 @@ export class ChatItemCard {
...
+      this.cardHeader.remove();
+      this.cardHeader = null;
+    }
+    if (this.props.chatItem.header != null) {
+      this.cardHeader = this.getCardHeader();
+      this.card?.render.insertChild('beforeend', this.cardHeader);
+
+      /**
+       * Generate header if available
+       */
```
francescoopiccoli, line 216: nit: why is this comment indented?

src/components/chat-item/chat-item-card.ts
```diff
@@ -190,6 +200,39 @@ export class ChatItemCard {
       this.initialSpinner?.[0]?.remove();
     }
 
+    /**
+     * Clear header block
+     */
```
francescoopiccoli, line 205: do you think it could make sense to encapsulate the clear logic and the generate logic into one or two functions? or is this the only place where we will use this logic?

francescoopiccoli (approved):
LGTM! Should we add e2e tests?

src/components/chat-item/chat-item-card.ts
```diff
@@ -101,6 +104,13 @@ export class ChatItemCard {
     });
   };
 
+  private readonly getCardHeader = (): ExtendedHTMLElement => {
+    return DomBuilder.getInstance().build({
+      type: 'div',
+      classNames: [ 'mynah-chat-item-card-header', 'mynah-card-inner-order-5' ]
```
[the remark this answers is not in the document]
Jurredr: Suggested by Dogus. The lower the order, the earlier it shows up.

src/components/chat-item/chat-item-card.ts
```diff
@@ -190,6 +200,39 @@ export class ChatItemCard {
...
+      this.cardHeader = null;
+    }
+    if (this.props.chatItem.header != null) {
+      this.cardHeader = this.getCardHeader();
+      this.card?.render.insertChild('beforeend', this.cardHeader);
+
+      /**
+       * Generate header if available
+       */
+      if (this.header != null) {
```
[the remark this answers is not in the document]
Jurredr: This is the convention following our eslint rules.

src/components/chat-item/chat-item-card.ts
Jurredr: It's directly being used here. It's the convention we use throughout this entire file for the generation, so I just followed that.

PR #239, opened 2025-02-18
Jurredr/falcon context arrow

Jurredr:
## Added
- A right arrow icon now show up on the most right side of the context items if they have children, and if the item is hovered / focused.
- A new `onContextSelected` event, is added, including the ability to prevent adding a context item to the prompt input text.

## License

### DIFF
--- a/docs/PROPERTIES.md
+++ b/docs/PROPERTIES.md
@@ -45,6 +45,9 @@ export interface MynahUIProps {
   onTabAdd?: (
     tabId: string,
     eventId?: string) => void;
+  onContextSelected?: (
+    contextItem: QuickActionCommand,
+  ) => boolean;
   onTabRemove?: (
     tabId: string,
     eventId?: string) => void;
@@ -507,6 +510,26 @@ This event will be fired when user clicks the add tab button or double clicks to
 onTabAdd?: (tabId: string):void => {
       console.log(`New tabId: ${tabId}`);
     };
+...
+```
+
+---
+
+### `onContextSelected`
+
+This event will be fired whenever a user selects an item from the context (`@`) list either using mouse click or keyboard actions. It is only triggered for items without children, i.e. only leaves in the tree. The data of the selected context item can be accessed through the `contextItem`. This event handler expects a boolean return:
+- Returning `true` indicates that the context item should be added to the prompt input text.
+- Returning `false` indicates that nothing should be added to the prompt input, and the triggering string should be cleared. E.g. if a user types `@wor` and presses tab on the `@workspace` action, the `@wor` would be removed from the prompt input and no context item will be added.
+
+```typescript
+...
+onContextSelected(contextItem: QuickActionCommand) {
+  if (contextItem.command === 'add_prompt') {
+    Log('Custom context action triggered for adding a prompt!')
+    return false;
+  }
+  return true;
+},
 ...
 ```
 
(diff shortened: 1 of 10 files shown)


docs/PROPERTIES.md
```diff
@@ -512,6 +515,26 @@ onTabAdd?: (tabId: string):void => {
 
 ---
 
+### `onContextSelected`
+
+This event will be fired whenever a user selects an item from the context (`@`) list either using mouse click or keyboard actions. It is only triggered for items without children, i.e. only leaves in the tree. The data of the selected context item can be accessed through the `contextItem`. This event handler expects a boolean return:
+- Returning `true` indicates that the context item should be added to the prompt input text.
+- Returning `false` indicates that nothing should be added to the prompt input, and the triggering string should be cleared. E.g. if a user types `@wor` and presses tab on the `@workspace` action, the `@wor` would be removed from the prompt input and no context item will be added.
```
francescoopiccoli, line 522: will there be any visual cue / UI element indicating the user that the context has been effectively selected in this 'false' scenario

src/components/chat-item/chat-prompt-input.ts
```diff
@@ -524,9 +524,7 @@ export class ChatPromptInput {
     }
   };
 
-  private readonly handleContextCommandSelection = (
-    contextCommand: QuickActionCommand,
-    route?: []): void => {
```
francescoopiccoli, line 529: just to double check, the removal of `route?: []` is intentional?

example/src/config.ts
```diff
@@ -300,6 +300,11 @@ export const mynahUIDefaults: Partial<MynahUITabStoreTab> = {
                     command: 'javascript_expert',
                     icon: MynahIcons.CHAT,
                     description: 'Expert on Javascript and typescript'
+                  },
+                  {
+                    command: 'add_prompt',
+                    icon: MynahIcons.PLUS,
+                    description: 'Add a new prompt'
```
francescoopiccoli, line 307: I dont understand why this was added. To test the `onContextSelected`  `false` scenario? If so why the naming `add_prompt `?

src/components/chat-item/chat-prompt-input.ts
```diff
@@ -538,10 +536,19 @@ export class ChatPromptInput {
...
       this.quickPick.close();
+      MynahUIGlobalEvents.getInstance().dispatch(MynahEventNames.CONTEXT_SELECTED, {
+        contextItem: contextCommand,
+        promptInputCallback: (insert: boolean) => {
+          if (insert) {
+            this.promptTextInput.insertContextItem({
+              ...contextCommand,
+            }, this.quickPickTriggerIndex);
+          } else {
+            this.promptTextInput.deleteTextRange(this.quickPickTriggerIndex, this.promptTextInput.getCursorPos());
```
francescoopiccoli, line 548: let's make sure to test this on different IDEs as well to be sure it works as expected

francescoopiccoli (approved):

docs/PROPERTIES.md
Jurredr: No, the idea is that the functionality of what happens after the selection is fully up to whoever implements the event handler. The context selector does always close on selection though, so that's a good indicator on its own.

example/src/config.ts
Jurredr: Yeah I get that, you'd need some more context for it. This 'add prompt' button is the whole reason why this `false` flag was implemented. I'll elaborate on Slack.

docs/PROPERTIES.md
francescoopiccoli: Alright, I think it could still be misleading or not super clear from a UX perspective if the context selector just closes, but I understand it's up to whoever implements the handler. Thanks for the clarification

PR #240, opened 2025-02-18
EXPERIMENTAL: file tree with overflow-y scroll instead of truncation

Jurredr:
## Problem
Allow file trees to scroll through overflow horizontally, instead of truncating text. As a preparation for larger file trees. Not sure if this is the way to go, and also not thoroughly tested, so highly experimental for now.

https://github.com/user-attachments/assets/e029a969-ffdc-47f7-aab1-591d5805e3d2

## Tests

## License

### DIFF
--- a/src/styles/components/chat/_chat-item-tree-view.scss
+++ b/src/styles/components/chat/_chat-item-tree-view.scss
@@ -73,7 +73,6 @@
             }
 
             .mynah-chat-item-pull-request-item {
-                // width: 100%;
                 box-sizing: border-box;
             }
         }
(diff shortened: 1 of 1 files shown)


src/styles/components/chat/_chat-item-tree-view.scss
```diff
@@ -77,7 +73,7 @@
             }
 
             .mynah-chat-item-pull-request-item {
-                width: 100%;
+                // width: 100%;
```
francescoopiccoli, line 76: does this need to be commented or should we just delete the line?

francescoopiccoli (approved):

src/styles/components/chat/_chat-item-tree-view.scss
Jurredr: Ah nice catch, will remove.

PR #242, opened 2025-02-19
Fixed form validator bug; not taking non-mandatory empty fields into account

Jurredr:
## Problem
The newly introduced form input validators should only apply if the field is non-empty or mandatory.

## Solution
Updated the text input validator function to take into account these parameters.

## License

### DIFF
--- a/src/helper/validator.ts
+++ b/src/helper/validator.ts
@@ -28,14 +28,13 @@ export const isTextualFormItemValid = (value: string, validationPatterns: {
       return prevValidation || isCurrentPatternValid;
     }, validationPatterns.operator === 'and');
   }
+  // Don't invalidate if the field is empty and non mandatory
+  isValid = isValid || ((value === undefined || value.trim() === '') && (mandatory !== true));
   if (isValid) {
     validationErrors = [];
   } else if (validationErrors.length === 0 && validationPatterns.genericValidationErrorMessage != null) {
     validationErrors.push(validationPatterns.genericValidationErrorMessage);
   }
-
-  // Don't invalidate if the field is empty and non mandatory
-  isValid = isValid || ((value === undefined || value.trim() === '') && (mandatory !== true));
   return { isValid, validationErrors };
 };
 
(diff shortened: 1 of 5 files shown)


src/helper/validator.ts
```diff
@@ -32,7 +33,25 @@ export const isTextualFormItemValid = (value: string, validationPatterns: {
   } else if (validationErrors.length === 0 && validationPatterns.genericValidationErrorMessage != null) {
     validationErrors.push(validationPatterns.genericValidationErrorMessage);
   }
+
+  // Don't invalidate if the field is empty and non mandatory
+  isValid = isValid || ((value === undefined || value.trim() === '') && (mandatory !== true));
   return { isValid, validationErrors };
```
francescoopiccoli, line 39: with the new change there could be a scenario where `isValid` is `false` and `validationErrors` is non empty right? for instance if there are validation errors but the field is non empty? wouldnt that be weird? 
we could either do the check before checking the validation patterns or by resetting the `validationErrors` in case the newly added logic for `isValid` returns true? Wdyt?

francescoopiccoli (approved):

src/helper/validator.ts
```diff
@@ -32,7 +33,25 @@ export const isTextualFormItemValid = (value: string, validationPatterns: {
   } else if (validationErrors.length === 0 && validationPatterns.genericValidationErrorMessage != null) {
     validationErrors.push(validationPatterns.genericValidationErrorMessage);
   }
+
+  // Don't invalidate if the field is empty and non mandatory
+  isValid = isValid || ((value === undefined || value.trim() === '') && (mandatory !== true));
```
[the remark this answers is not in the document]
Jurredr: This is the intended behavior, correct. Undefined should indeed evaluate to true

src/helper/validator.ts
Jurredr: Fair point, my quick work around was to no longer check for the amount of `validationErrors`, but just to check for `isValid` for validity checking. It is indeed a bit of a weird pattern. I'll modify it so that `validationErrors` is emptied if `isValid` is false.

PR #245, opened 2025-02-24
Jurredr/flat file list

Jurredr:
## Added
Added a prop to render a flat file list for file tree component.

## Tests

## License

### DIFF
--- a/src/components/chat-item/chat-item-card.ts
+++ b/src/components/chat-item/chat-item-card.ts
@@ -337,7 +337,7 @@ export class ChatItemCard {
       this.fileTreeWrapper = null;
     }
     if (this.props.chatItem.fileList != null) {
-      const { filePaths = [], deletedFiles = [], actions, details } = this.props.chatItem.fileList;
+      const { filePaths = [], deletedFiles = [], actions, details, flatList } = this.props.chatItem.fileList;
       const referenceSuggestionLabel = this.props.chatItem.body ?? '';
       this.fileTreeWrapper = new ChatItemTreeViewWrapper({
         tabId: this.props.tabId,
@@ -349,6 +349,7 @@ export class ChatItemCard {
         collapsedByDefault: this.props.chatItem.fileList.collapsedByDefault ?? false,
         files: filePaths,
         deletedFiles,
+        flatList,
         actions,
         details,
         references: this.props.chatItem.codeReference ?? [],
--- a/src/components/chat-item/chat-item-tree-view-wrapper.ts
+++ b/src/components/chat-item/chat-item-tree-view-wrapper.ts
@@ -57,13 +57,12 @@ export class ChatItemTreeViewWrapper {
         node: fileListToTree(props.files, props.deletedFiles, props.actions, props.details, props.rootTitle),
         hideFileCount: props.hideFileCount,
         collapsedByDefault: props.collapsedByDefault,
-        flatList: props.flatList
       }).render;
 
     this.render = DomBuilder.getInstance().build({
       type: 'div',
       testId: testIds.chatItem.fileTree.wrapper,
-      classNames: [ 'mynah-chat-item-tree-view-wrapper', ...(props.classNames ?? []) ],
+      classNames: [ 'mynah-chat-item-tree-view-wrapper', props.flatList === true ? 'mynah-chat-item-tree-view-flat-list' : '', ...(props.classNames ?? []) ],
       children: [
         {
           type: 'div',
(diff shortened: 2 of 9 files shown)


src/helper/file-tree.ts
```diff
@@ -58,18 +58,56 @@ export const fileListToTree = (
...
+    });
+
+    return rootNode;
+  }
+
+  // Original hierarchical tree logic
   return [ ...splitFilePaths(modifiedFilePaths, false), ...splitFilePaths(deletedFilePaths, true) ].reduce<TreeNode>(
     (acc, { originalFilePath, filePath, deleted }) => {
-      // pointer to keep track of the current tree node
+      // ... existing code ...
```
dogusata, line 105: ?

src/helper/file-tree.ts
```diff
@@ -58,18 +58,56 @@ export const fileListToTree = (
   deletedFilePaths: string[] = [],
   actions?: Record<string, FileNodeAction[]>,
   details?: Record<string, TreeNodeDetails>,
-  rootTitle?: string): TreeNode => {
+  rootTitle?: string,
+  flatList: boolean = false): TreeNode => {
```
dogusata, line 62: In general nice work! **However**, I'm not sure if we really need this much effort and js workaround to make them look like flat? in `_chat-item-tree-view.scss` line 60, with a parent css selector we can flat them out and hide folders except the root folder very easily. With just a simple css toggle, we also unlock the possibility of a flat-tree view toggle in the `fileTree` component in the future in case we expose it as a standalone component too. 

Also, while doing it just with css, it makes sense to rename the weird `mynah-chat-item-pull-request-item` selector to a properly named one too.

dogusata (changes requested):

src/components/chat-item/chat-item-card.ts
```diff
@@ -349,6 +349,7 @@ export class ChatItemCard {
         collapsedByDefault: this.props.chatItem.fileList.collapsedByDefault ?? false,
         files: filePaths,
```
dogusata, line 350: any reason not to `...this.props.chatItem.fileList`?

dogusata: If it is just the files≠fileList, I guess we can update the ChatItemTreeWrapper component.

src/components/chat-item/chat-item-tree-view.ts
```diff
@@ -74,7 +78,7 @@ export class ChatItemTreeView {
     const folderItem = new Button({
       testId: testIds.chatItem.fileTree.folder,
       icon: new Icon({ icon: this.isOpen ? MynahIcons.DOWN_OPEN : MynahIcons.RIGHT_OPEN }).render,
-      classNames: [ 'mynah-chat-item-tree-view-button' ],
+      classNames: [ 'mynah-chat-item-tree-view-button', this.depth === 0 ? 'root' : '' ],
```
dogusata, line 81: I'm not sure if it is necessary, it is possible to understand it through css. And even if it not and you prefer using an additional class, always be specific to avoid overrides for generic names. `root` is very generic.

src/styles/components/chat/_chat-item-tree-view.scss
```diff
@@ -57,11 +57,19 @@
 
             > .mynah-chat-item-tree-view-button {
                 gap: var(--mynah-sizing-1);
-                & ~ .mynah-chat-item-pull-request-item {
+                & ~ .mynah-chat-item-folder-child:not(.flat-list),
+                &.root ~ .mynah-chat-item-folder-child {
```
dogusata, line 61: I guess we can use the `flat-list` only on the top level, the rest can be handled through children/siblink selectors.

src/components/chat-item/chat-item-tree-view-wrapper.ts
```diff
@@ -55,7 +56,8 @@ export class ChatItemTreeViewWrapper {
         tabId: props.tabId,
         node: fileListToTree(props.files, props.deletedFiles, props.actions, props.details, props.rootTitle),
         hideFileCount: props.hideFileCount,
-        collapsedByDefault: props.collapsedByDefault
+        collapsedByDefault: props.collapsedByDefault,
+        flatList: props.flatList
```
dogusata, line 60: I guess having `flat-list` only on the top level (this component) is enough, the rest can be handled in css through children/sibling selectors.

dogusata (changes requested):

src/components/chat-item/chat-item-tree-view.ts
Jurredr: It might be possible with pure css, but I don't see the point in making my life harder haha

dogusata (approved):

PR #248, opened 2025-02-25
Dogusata beta/fix card header refresh

dogusata:
## Problem
- It is not possible to use the shift key during filtering the items when the context selector panel is open to add an underscore (`_`) into the search term. It closes the context selector panel when shift is pressed combined with any key
- When the action button on the right of the context group title clicked, the inserted `@` remains on the screen.
- Filtering in quick pick items does make the filter however it doesn't properly order, items contain the search term at the end of their text can appear at the top of the list. And it also shows some items doesn't match the term at all.
- Sometimes the context selection panel gets stuck and doesn't reflect to any input. It requires `esc` keypress to close it and reopen it back
- When choosing items from the context provider, the first item should be indicated as selected before navigating since if user hits enter the first item is being selected.
- When a rich content copied from outside chat panel is pasted to the prompt input, it keeps the styling inside the prompt input. It should only use the plain text.
- Right arrow on context items with children causes the description text to jump to next line and doesn't look vertically aligned in center.
- `header` is re-rendering every time the stream is updated. If the `header` attribute is not sent every time, it shouldn't update the render too
- fileList is insufficient, not able to remove folder icons or use another icon for them.
- Prompt input field doesn't seem like it is possible to insert multiline content. Also, if the placeholder is long, it doesn't show the whole text.

## Solution
- Shift and other modifier keys are not closing the panel. Previously when shift or other modifier keys were pressed, their key definition _(like `"SHIFT"`)_ were being added to search term and it was hiding the panel since there were no items available in the list instead of actually closing the panel.
- When an action button in command/context group title is clicked, it removes the given trigger char (`@` or `/`)
- Filtering in quick pick items is improved with complex match structure and a flat list instead of using groups for easy and better filtering. It can find most relevant items, order them according to their match score and more importantly it highlights the match in the text for clear visibility.
- Context selection panel closes as expected now, solved together with SHIFT or other modifier key presses.
- First item in the quick pick items list is now indicated as it is the target, so users can understand if they hit Enter that item will be selected.
- Rich contents are being converted to plain text before they're being pasted to prompt input.
- Right arrow and other elements on quick pick items are moved to a grid structure to better layout and rendering
- `header` block in ChatCard is not being re-rendered **if it is sent as filled once and than sent as `undefined` (at least once) specifically**. The reason behind that is to avoid re-rendering it. 
- added `folderIcon` option to `fileList` as `MynahIcons | MynahIconsType | null`. If `null` is sent, it will not show icons for folders including the root folder.
- Prompt input min-height is increased to 3 lines height **only** if the window height is **bigger than 450px**.

#### Header (fileList)
image

#### Filter
image

#### Prompt input height
image

## Tests

## License

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@aws/mynah-ui",
     "displayName": "AWS Mynah UI",
-    "version": "4.23.0-beta.10",
+    "version": "4.23.0-beta.11",
     "description": "AWS Toolkit VSCode and Intellij IDE Extension Mynah UI",
     "publisher": "Amazon Web Services",
     "license": "Apache License 2.0",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@aws/mynah-ui",
     "displayName": "AWS Mynah UI",
-    "version": "4.23.0-beta.9",
+    "version": "4.23.0-beta.10",
     "description": "AWS Toolkit VSCode and Intellij IDE Extension Mynah UI",
     "publisher": "Amazon Web Services",
     "license": "Apache License 2.0",
--- a/src/components/chat-item/chat-item-card.ts
+++ b/src/components/chat-item/chat-item-card.ts
@@ -200,36 +200,37 @@ export class ChatItemCard {
       this.initialSpinner?.[0]?.remove();
     }
 
-    /**
-     * Clear header block
-     */
-    if (this.cardHeader != null) {
-      this.cardHeader.remove();
-      this.cardHeader = null;
-    }
-    if (this.props.chatItem.header != null) {
-      this.cardHeader = this.getCardHeader();
-      this.card?.render.insertChild('beforeend', this.cardHeader);
-
-      /**
-       * Generate header if available
-       */
-      if (this.header != null) {
-        this.header.render.remove();
-        this.header = null;
+    // If no data is provided for the header
+    // skip removing and checking it
+    if (this.props.chatItem.header !== undefined) {
+      if (this.cardHeader != null) {
+        this.cardHeader.remove();
+        this.cardHeader = null;
       }
       if (this.props.chatItem.header != null) {
-        this.header = new ChatItemCard({
-          tabId: this.props.tabId,
-          small: true,
-          inline: true,
-          chatItem: {
-            ...this.props.chatItem.header,
-            type: ChatItemType.ANSWER,
-            messageId: this.props.chatItem.messageId
-          }
-        });
-        this.cardHeader.insertChild('beforeend', this.header.render);
+        this.cardHeader = this.getCardHeader();
+        this.card?.render.insertChild('beforeend', this.cardHeader);
+
+        /**
+         * Generate header if available
+         */
+        if (this.header != null) {
+          this.header.render.remove();
+          this.header = null;
+        }
+        if (this.props.chatItem.header != null) {
+          this.header = new ChatItemCard({
+            tabId: this.props.tabId,
+            small: true,
+            inline: true,
+            chatItem: {
+              ...this.props.chatItem.header,
+              type: ChatItemType.ANSWER,
+              messageId: this.props.chatItem.messageId
+            }
+          });
+          this.cardHeader.insertChild('beforeend', this.header.render);
+        }
       }
     }
 
--- a/src/components/chat-item/chat-item-card.ts
+++ b/src/components/chat-item/chat-item-card.ts
@@ -346,6 +346,7 @@ export class ChatItemCard {
         messageId: this.props.chatItem.messageId ?? '',
         cardTitle: this.props.chatItem.fileList.fileTreeTitle,
         rootTitle: this.props.chatItem.fileList.rootFolderTitle,
+        folderIcon: this.props.chatItem.fileList.folderIcon,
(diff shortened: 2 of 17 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "@aws/mynah-ui",
     "displayName": "AWS Mynah UI",
-    "version": "4.23.0-beta.9",
+    "version": "4.23.0-beta.11",
```
Jurredr, line 4: We're skipping 10? 😛

src/components/chat-item/chat-item-card.ts
```diff
@@ -200,36 +200,37 @@ export class ChatItemCard {
...
-
-      /**
-       * Generate header if available
-       */
-      if (this.header != null) {
-        this.header.render.remove();
-        this.header = null;
+    // If no data is provided for the header
+    // skip removing and checking it
+    if (this.props.chatItem.header !== undefined) {
```
Jurredr, line 205: Nice, as we discussed, the generation code for different components in the ChatItemCard can definitely use some refinement. Not necessary for now, but I think there is some good room to clear up the code in this file.

src/components/chat-item/chat-prompt-input.ts
```diff
@@ -524,16 +530,17 @@ export class ChatPromptInput {
...
+    const contextCommand = {
+      ...dirtyContextCommand,
+      command: dirtyContextCommand.command.replace(new RegExp(`${MARK_OPEN}|${MARK_CLOSE}`, 'g'), '')
+    };
     // Check if the selected command has children
-    const children = this.quickPickItemGroups
-      .flatMap(group => group.commands)
-      .find(cmd => cmd.command === contextCommand.command)?.children;
-    if (children != null && children.length > 0) {
+    if (contextCommand.children != null && contextCommand.children?.length > 0) {
```
Jurredr, line 539: Small nit; optional chaining on `children` not necessary after null check

src/components/chat-item/chat-prompt-input.ts
```diff
@@ -524,16 +530,17 @@ export class ChatPromptInput {
     }
   };
 
-  private readonly handleContextCommandSelection = (contextCommand: QuickActionCommand): void => {
+  private readonly handleContextCommandSelection = (dirtyContextCommand: QuickActionCommand): void => {
+    const contextCommand = {
+      ...dirtyContextCommand,
+      command: dirtyContextCommand.command.replace(new RegExp(`${MARK_OPEN}|${MARK_CLOSE}`, 'g'), '')
```
Jurredr, line 536: Nit; duplicate code for the `RegExp`. Probably better to extract to a constant?

src/components/chat-item/prompt-input/prompt-text-input.ts
```diff
@@ -80,12 +84,39 @@ export class PromptTextInput {
...
             this.props.onBlur();
           }
-        }
+        },
+        paste: (e: ClipboardEvent) => {
+          // Prevent the default paste behavior
+          e.preventDefault();
+
+          // Get plain text from clipboard
+          const text = e.clipboardData?.getData('text/plain') ?? '';
```
Jurredr, line 99: If this text is null, shouldn't we just return?

src/components/chat-item/prompt-input/prompt-text-input.ts
```diff
@@ -342,23 +391,12 @@ export class PromptTextInput {
...
-    if ((selection != null) && selection.rangeCount !== 0) {
-      const range = selection.getRangeAt(0);
-      const preCaretRange = range.cloneRange();
-      preCaretRange.selectNodeContents(this.promptTextInput);
-      preCaretRange.setEnd(range.endContainer, range.endOffset);
-      position = preCaretRange.toString().length;
-    }
-
-    return position;
+    this.checkIsEmpty();
```
Jurredr, line 395: nice!

src/helper/quick-pick-data-handler.ts
```diff
@@ -1,31 +1,141 @@
...
+    } ];
+  }
+
+  return [];
+};
+
+export const MARK_OPEN = '<mark>';
+export const MARK_CLOSE = '</mark>';
+
+const highlightMatch = (text: string, searchTerm: string): string => {
```
Jurredr, line 49: Does this actually highlight the match? Or does it `find` the best match? If so, perhaps better to rename to `findBestMatch`

src/components/chat-item/chat-item-card.ts
dogusata: We'll create a new flow for that. The reason i see is not just related with this check, but also related with preparation  of this data. Even though the consumer doesn't send it (not specifically undefined but not just sending it) we're spreading the previously set data first. Which means that if you give it once, it will be used every time which will cause the re-rendering.

src/helper/quick-pick-data-handler.ts
dogusata: it highlights the exact match and also the fuzzy match with chars or char groups too. And this one really highlights them. Just to wrap them with `<mark></mark>` items. Find is a different function.

src/components/chat-item/chat-prompt-input.ts
dogusata: Applied a different way, since length doesn't work but `children?.[0]` works.

Jurredr (approved):

PR #259, opened 2025-03-04
Jurredr/context description truncate

Jurredr:
## Problem
- Context item descriptions could be too long, causing overflow and lack of readability. 

## Solution
Truncate with ellipsis in the middle.
Screenshot 2025-03-04 at 14 23 04
Screenshot 2025-03-04 at 14 22 58

## Tests

## License

### DIFF
--- a/src/components/chat-item/prompt-input/prompt-input-quick-pick-item.ts
+++ b/src/components/chat-item/prompt-input/prompt-input-quick-pick-item.ts
@@ -12,8 +12,24 @@ export interface PromptInputQuickPickItemProps {
 export class PromptInputQuickPickItem {
   render: ExtendedHTMLElement;
   private readonly props: PromptInputQuickPickItemProps;
+  private readonly descriptionText: ExtendedHTMLElement;
+  private readonly description: ExtendedHTMLElement;
+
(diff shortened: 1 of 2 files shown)


dogusata (changes requested):

src/components/chat-item/prompt-input/prompt-input-quick-pick-item.ts
```diff
@@ -68,6 +74,8 @@ export class PromptInputQuickPickItem {
           : [])
       ]
     });
+
+    setTimeout(() => this.updateTruncation(), 5);
```
dogusata, line 78: What will happen if the element is not on screen yet? Delaying more until it renders on screen?

dogusata (changes requested):

dogusata:
A few more things as a general feedback:
- What will happen if the screen size changes? Will we recalculate? Why?
- It feels like that is too much calculation, and with a hidden element etc. Can we simplify it? 
- It can try to calculate some string before even actually the component is being rendered? Adding more delay wouldn't help on that. We're generally thinking that all components can be used in any place, this may not work in "any" place.

dogusata (approved):

PR #262, opened 2025-03-10
Dogusata/imrovements on styling

dogusata:
## Problem
- Animations are slower than they should be
- It is not possible to understand if there is no match but the panel is still open while filtering the quick pick items
- Progress bar looks so thick since the min height is 3 lines for the prompt input, it should be reduced
- Context items causing new lines in the user prompt chat item view (theme colored cards on the right)
- Context item description field middle ellipsis causes the character to be clipped from half sometimes.

## Solution
- Animations are reduced to half or complete flash
- We now have a search title on the top of the quick pick item selector
  - The search term followed by a search icon.
  - We now have option to add icons to quick pick item groups too
- Prompt input's min-height is being shrink down to 1 line if the progress is active
- Sent value for the prompt input is fixed to provide plain text with user intended new lines.
- Context item description is now being done through 2 tricks, it still cuts the text from the middle and uses two parts on left and right:
  - Left part is not being clipped, instead it breaks words to new lines char-by-char depending on line width. But the new lines are not visible because of max-height. Which makes the text clipped but without clipping the characters from their middle
  - The right part has an ellipsis, but on the beginning. To allow it it has dir=rtl. But when you use dir=rtl, it moves the end punctuation to the most left, to avoid it used <bdi> tag, which has enough support for the target browsers we have.

image
image

image
image

## Checks

## License


example/src/main.ts
```diff
@@ -185,6 +190,9 @@ export const createMynahUI = (initialData?: MynahUIDataModel): MynahUI => {
         }, 10000);
       } else if (buttonId === 'custom-data-check') {
         // Use for custom temporary checks
+        mynahUI.updateStore(tabId, {
```
Jurredr, line 193: Nit; clear the custom data check (this is just for quick temporary checks during development right?)

src/components/chat-item/prompt-input/prompt-input-quick-pick-item.ts
```diff
@@ -55,10 +56,10 @@ export class PromptInputQuickPickItem {
               classNames: [ 'mynah-chat-command-selector-command-description' ],
               children: [ {
                 type: 'span',
-                children: [ this.props.quickPickItem.description.slice(0, Math.ceil(this.props.quickPickItem.description.length / 2)) ]
+                innerHTML: this.props.quickPickItem.description.slice(0, descriptionSplitPosition)
               }, {
                 type: 'span',
-                children: [ this.props.quickPickItem.description.slice(Math.ceil(this.props.quickPickItem.description.length / 2)) ]
+                innerHTML: `<bdi>${this.props.quickPickItem.description.slice(descriptionSplitPosition)}</bdi>`
```
Jurredr, line 62: Very curious about this one, can you elaborate on `bdi`?

src/styles/components/chat/_chat-command-selector.scss
```diff
@@ -127,30 +136,35 @@
...
                 > span:last-child {
-                    flex: 0 1 auto;
-                    white-space: pre;
-                    overflow: hidden;
-                    display: inline-flex;
-                    flex-flow: row nowrap;
-                    justify-content: flex-end;
+                    display: inline-block;
+                    white-space: nowrap;
+                    direction: rtl;
```
Jurredr, line 157: using rtl again?

ui-tests/__test__/__image_snapshots__/chromium/main-spec-ts-open-mynah-ui-should-show-progress-indicator-1-snap.png
Jurredr, line 1: So progress bar now shrinks to 1 lineheight? Does that not look weird when transitioning from the regular chat input?

Jurredr (approved):

src/components/chat-item/prompt-input/prompt-input-quick-pick-item.ts
dogusata: Written in the PR details section already

src/styles/components/chat/_chat-command-selector.scss
dogusata: Yes, written in the PR description

ui-tests/__test__/__image_snapshots__/chromium/main-spec-ts-open-mynah-ui-should-show-progress-indicator-1-snap.png
dogusata: No, it does it smoothly

example/src/main.ts
dogusata: for the next one sure.

Jurredr (approved):

PR #265, opened 2025-03-11
Added basic data model for InformationItem

Jurredr:
## Added
A data model for `InformationItem` and `InformationItemGroup`.

```typescript
export interface InformationItemGroup {
  title?: string;
  icon?: MynahIcons | MynahIconsType;
  children: InformationItem[];
}

export interface InformationItem {
  messageId?: string;
  icon?: MynahIcons | MynahIconsType;
  title?: string;
  description?: string;
  actions?: ChatItemButton[];
  active?: boolean;
  clickable?: boolean;
}
```

- `chatItems?` array on a tab can now contain either `ChatItem` objects or `InformationItemGroup` objects
- `InformationItemGroup` can optionally have a title to show above the group
- `InformationItem` must have a title, but can optionally have a description, be set to active to add a subtle background color, have a number of actions rendered on the right, and can have an onClick action on the whole item

## Tests

## License

### DIFF
--- a/src/components/chat-item/chat-item-card.ts
+++ b/src/components/chat-item/chat-item-card.ts
@@ -18,7 +18,7 @@ import { ChatItemFormItemsWrapper } from './chat-item-form-items';
 import { ChatItemButtonsWrapper, ChatItemButtonsWrapperProps } from './chat-item-buttons';
 import { cleanHtml } from '../../helper/sanitize';
 import { CONTAINER_GAP } from './chat-wrapper';
-import { chatItemHasContent } from '../../helper/chat-item';
+import { chatItemHasContent, isChatItem } from '../../helper/chat-item';
 import { Card } from '../card/card';
 import { ChatItemCardContent, ChatItemCardContentProps } from './chat-item-card-content';
 import testIds from '../../helper/test-ids';
@@ -543,7 +543,9 @@ export class ChatItemCard {
             .getTabDataStore(this.props.tabId)
             .updateStore(
               {
-                chatItems: currentTabChatItems?.map((chatItem: ChatItem) => {
+                chatItems: currentTabChatItems?.filter((chatItem): chatItem is ChatItem =>
+                  isChatItem(chatItem)
+                ).map((chatItem: ChatItem) => {
                   if (chatItem.messageId === this.props.chatItem.messageId) {
                     return this.props.chatItem;
                   }
--- a/src/helper/serialize-chat.ts
+++ b/src/helper/serialize-chat.ts
@@ -1,5 +1,7 @@
 import { ChatItemCard } from '../components/chat-item/chat-item-card';
 import { MynahUITabsStore } from './tabs-store';
+import { ChatItem } from '../static';
+import { isChatItem } from './chat-item';
 
 /**
  * Serialize all (non-empty) chat messages in a tab to a markdown string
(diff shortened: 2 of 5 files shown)


src/static.ts
```diff
@@ -222,6 +222,21 @@ export enum ChatItemType {
   CODE_RESULT = 'code-result',
 }
 
+export interface InformationItemGroup {
+  title?: string;
+  informationItems: InformationItem[];
```
dogusata, line 227: I believe just `children` should be better and more generic. In case if we also want to let it render different items in the future. Like we're missing on the `chatItems` part. Actually we'll render something different now than a chat item but for backwards compatibility we're using it as is. Most likely we'll also rename it to something else with the V5.

src/static.ts
```diff
@@ -222,6 +222,21 @@ export enum ChatItemType {
   CODE_RESULT = 'code-result',
 }
 
+export interface InformationItemGroup {
+  title?: string;
```
dogusata, line 226: To better extend it with `icon?: MynahIcons | MynahIconsType` and `description?:string` would be nice.

src/static.ts
```diff
@@ -222,6 +222,21 @@ export enum ChatItemType {
...
+}
+
+export interface InformationItem {
+  messageId?: string;
+  icon?: MynahIcons | MynahIconsType;
+  title: string;
+  description?: string;
+  actions?: ChatItemButton[];
+  active?: boolean;
+  onClick?: () => void;
```
dogusata, line 237: We cannot use/define function types in the data model. The reason behind that they are not serializable. It will break the ideal persistency approach. `data` model should be serializable. But if we use function types (or window etc.) it won't be serializable anymore and we cannot convert the data to a string. If we use any kind of unserializable object type (function, window etc.) it will also break the current behavior when we start to use it. 

Instead we need to expose a new event maybe called `onInformationItemClick` on main class properties and put here a boolean option only like `clickable?:boolean`.

src/static.ts
```diff
@@ -222,6 +222,21 @@ export enum ChatItemType {
...
 
+export interface InformationItemGroup {
+  title?: string;
+  informationItems: InformationItem[];
+}
+
+export interface InformationItem {
+  messageId?: string;
+  icon?: MynahIcons | MynahIconsType;
+  title: string;
```
dogusata, line 233: I believe title can also be optional. In some cases we may only want to use the `description`. Let it decided by the consumer, I guess it is not necessary to push consumer to provide a title here.

dogusata (changes requested):

src/components/chat-item/chat-item-card.ts
```diff
@@ -543,7 +543,9 @@ export class ChatItemCard {
             .getTabDataStore(this.props.tabId)
             .updateStore(
               {
-                chatItems: currentTabChatItems?.map((chatItem: ChatItem) => {
+                chatItems: currentTabChatItems?.filter((chatItem): chatItem is ChatItem =>
+                  isChatItem(chatItem)
```
dogusata, line 547: Is it just to avoid the usage for now? Since we're expecting to render them without grouping (I mean by grouping; grouping and rendering the chat items in a block and the InformationItemGroups in another block). They can be used mixed.

src/helper/serialize-chat.ts
```diff
@@ -1,13 +1,17 @@
...
 
 /**
  * Serialize all (non-empty) chat messages in a tab to a markdown string
  * @param tabId Corresponding tab ID.
  * @returns The bodies of chat cards in markdown format, separated by \n\n---\n\n
  */
 export const serializeMarkdown = (tabId: string): string => {
-  return MynahUITabsStore.getInstance().getAllTabs()[tabId].store?.chatItems?.map(chatItem => chatItem.body ?? '').filter(chatItem => chatItem.trim() !== '').join('\n\n---\n\n') ?? '';
+  return MynahUITabsStore.getInstance().getAllTabs()[tabId].store?.chatItems?.filter((chatItem): chatItem is ChatItem =>
+    isChatItem(chatItem)
```
dogusata, line 13: Why we're not adding the information items to export too?

src/helper/serialize-chat.ts
```diff
@@ -16,7 +20,9 @@ export const serializeMarkdown = (tabId: string): string => {
  * @returns The bodies of chat cards in HTML format
  */
 export const serializeHtml = (tabId: string): string => {
-  const chatItemCardDivs = MynahUITabsStore.getInstance().getAllTabs()[tabId].store?.chatItems?.filter(chatItem => (chatItem.body != null) && chatItem.body.trim() !== '').map(chatItem => new ChatItemCard({
+  const chatItemCardDivs = MynahUITabsStore.getInstance().getAllTabs()[tabId].store?.chatItems?.filter((chatItem): chatItem is ChatItem =>
+    isChatItem(chatItem) && chatItem.body != null && chatItem.body.trim() !== ''
```
dogusata, line 24: same here

dogusata (approved):

PR #266, opened 2025-03-13
Jurredr/form input icons events

Jurredr:
## Added
- `prefixIcon` for all form items except for stars and radio group, an icon showing at the beginning:

image

## Changed
- All global event dispatches from `chat-item-form-items` have been moved to callback handlers
- Added `onFormChange` event which is called every time anything in a form changes, regardless of validity

## Tests

## License

### DIFF
--- a/src/components/form-items/select.ts
+++ b/src/components/form-items/select.ts
@@ -5,7 +5,7 @@
 
 import { Config } from '../../helper/config';
 import { DomBuilder, DomBuilderObject, ExtendedHTMLElement } from '../../helper/dom';
-import { Icon, MynahIcons } from '../icon';
+import { Icon, MynahIcons, MynahIconsType } from '../icon';
 import '../../styles/components/_form-input.scss';
 
 interface SelectOption {
@@ -16,7 +16,8 @@ interface SelectOption {
 export interface SelectProps {
   classNames?: string[];
   attributes?: Record<string, string>;
-  icon?: MynahIcons;
+  icon?: MynahIcons | MynahIconsType;
+  prefixIcon?: MynahIcons | MynahIconsType;
   label?: HTMLElement | ExtendedHTMLElement | string;
   description?: ExtendedHTMLElement;
   value?: string;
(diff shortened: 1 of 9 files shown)


src/components/form-items/select.ts
```diff
@@ -16,7 +16,8 @@ interface SelectOption {
 export interface SelectProps {
   classNames?: string[];
   attributes?: Record<string, string>;
-  icon?: MynahIcons;
+  icon?: MynahIcons | MynahIconsType;
```
dogusata, line 19: What's the difference? Why we have two icons?

src/styles/components/_form-input.scss
```diff
@@ -176,11 +181,15 @@
...
+        > .mynah-form-input-icon {
+            color: var(--mynah-color-text-weak);
+            margin-right: var(--mynah-sizing-2);
+        }
         > .mynah-select-handle {
             position: absolute;
             color: var(--mynah-color-text-weak);
             pointer-events: none;
-            transform: translateX(-100%);
+            transform: translateX(-25%);
```
dogusata, line 192: Any specific reason for that? Did we check very long and short widths are still working with -25%?

dogusata:
And as a general comment, to keep it consistent across the whole mode, please use `icon` instead of `prefixIcon`. Since we don't support more than one icon spot (for almost nothing) it is better to say just `icon`.

src/components/form-items/select.ts
Jurredr: I thought this was our new standard? For compatibility with VSCode which was having some issues with the Icons type?

dogusata (approved):

ui-tests/__test__/__image_snapshots__/chromium/main-spec-ts-open-mynah-ui-forms-should-disable-forms-on-submit-1-snap.png
dogusata, line 1: This needs to be reverted, it seems like the test is failing and we're overriding with the failed snapshot. Previous one is correct, not the new one.

dogusata (changes requested):

dogusata (approved):

Issue #268, opened 2025-03-14
It's possible to add attributes to `updateStore` that are not part of `MynahUIDataModel`, potentially breaking store updates


kmile:
## Steps to reproduce the issue

Call updateStore:

```
type ExtendedDataModel = MynahUIDataModel & { someOtherProperty: boolean }
const props: ExtendedDataModel = { someOtherProperty: true }
mynahUi.updateStore(tabId, props)
```

This causes [updateStore](https://github.com/aws/mynah-ui/blob/main/src/helper/store.ts#L123) to break since there are no subscribers defined for `someOtherProperty`.

## Expected behavior

Additional properties are ignored when triggering subscribers.

PR #269, opened 2025-03-14
fixed: data store, defaults and subscriber extensibility

dogusata:
## Problem
It is possible to extend data store, defaults and subscribers event though they are typed.

## Solution
Checking if the incoming key is available in the actual data model.

## Tests

## License


Jurredr (dismissed):

kmile (dismissed):
Could we add a test case that uses the pattern from the issue #268 to verify this fix?

dogusata:
> Could we add a test case that uses the pattern from the issue #268 to verify this fix?

I guess it would be better to release this first, since it doesn't cause any problems on the current flows as tests are passing. We need to add a whole test suite (unit) for this  functionality imo. Which we start to work next week

PR #271, opened 2025-03-19
Added sheet component, simplified feedback form to consume sheet

Jurredr:
## Added / Changed
- Added the `Sheet` component to abstract the custom form's display for other uses.
  - It can be set to `fullScreen: true`, which makes it take up the full screen without any borders.
  - The `feedback-form` now calls to the `sheet` under the hood, passing along its chat items to propagate the sheet body.
  - The `custom-form` was removed due to redundancy and is now merged into the `feedback-form`.
- Added the `DetailedListSheet`, a wrapper for a sheet with a `DetailedList` inside of it.
- Added the `mynahUI.openDetailedList()` public function for opening such a `DetailedListSheet`.

#### Sheet
```typescript
export interface SheetProps {
  tabId: string;
  title?: string;
  children?: Array<ExtendedHTMLElement | HTMLElement | string | DomBuilderObject>;
  fullScreen?: boolean;
  description?: string;
  onClose: () => void;
}
```
#### DetailedListSheet
```typescript
const { update, close, changeTarget, getTargetElementId } = mynahUI.openDetailedList({
    tabId,
    detailedList: {}, // The DetailedList, for the data model check DATAMODEL.md
    events: {
      onFilterValueChange: (filterValues: Record<string, any>, isValid: boolean) => {
      },
      onKeyPress: (e) => {
      },
      onItemSelect: (detailedListItem) => {
      },
      onActionClick: (button) => {
      },
      onClose: () => {
      },
    }
});          
```

#### Sheet component
Screenshot 2025-04-01 at 15 03 21

#### Full screen sheet
Screenshot 2025-04-01 at 15 03 21

#### Examples of DetailedListSheet 
Screenshot 2025-04-01 at 15 03 33
Screenshot 2025-04-01 at 15 03 21

## Tests

## License


src/components/detailed-list/detailed-list-item.ts
```diff
@@ -44,16 +51,20 @@ export class DetailedListItemWrapper {
               }
             ]
           : []),
-        {
-          type: 'div',
-          classNames: [ 'mynah-detailed-list-item-name' ],
-          innerHTML: this.props.listItem.title ?? this.props.listItem.name
-        },
+        ...(this.props.listItem.title !== undefined || this.props.listItem.name !== undefined
```
dogusata, line 54: just `!=` should be enough. This is not the case but sometimes to reset items, `delete` would basically not work because of immutable property, but we can set them to `null`. So `!=null` would do the trick for both `undefined` and `null`.

src/components/detailed-list/detailed-list-item.ts
```diff
@@ -83,4 +115,47 @@ export class DetailedListItemWrapper {
...
   };
+
+  private readonly showActionMenuOverlay = (): void => {
+    this.actionMenuOverlay = new Overlay({
+      background: true,
+      closeOnOutsideClick: true,
+      referenceElement: this.render,
+      dimOutside: false,
+      removeOtherOverlays: true,
+      verticalDirection: OverlayVerticalDirection.START_TO_BOTTOM,
```
dogusata, line 126: Since we're aligning these actions always on right, I guess it would be better to say `CENTER` to vertical and `LEFT` to horizontal.

src/components/detailed-list/detailed-list.ts
```diff
@@ -12,8 +12,9 @@ import { TitleDescriptionWithIcon } from '../title-description-with-icon';
 export interface DetailedListWrapperProps {
   detailedList: DetailedList;
   onFilterValueChange?: (filterValues: Record<string, any>, isValid: boolean) => void;
-  onDetailedListItemGroupActionClick: (action: ChatItemButton) => void;
-  onDetailedListItemSelect: (detailedListItem: DetailedListItem) => void;
+  onDetailedListItemGroupActionClick?: (action: ChatItemButton) => void;
```
dogusata, line 15: nit: Since those are part of the `DetailedList..` component itself already, i guess the event names can be simplified as:
```
onGroupActionClick
onItemActionClick
onItemSelect
```

ui-tests/__test__/flows/feedback-form/render-feedback-form.ts
```diff
@@ -34,8 +34,8 @@ export const renderFeedbackForm = async (page: Page, skipScreenshots?: boolean):
   await commentInput.fill('This is some feedback comment');
   await waitForAnimationEnd(page);
 
-  const descriptionText = page.locator(getSelector(testIds.feedbackForm.description));
-  expect(descriptionText).toBeDefined();
+  // const descriptionText = page.locator(getSelector(testIds.feedbackForm.description));
```
dogusata, line 37: Are those necessary comments?

src/styles/components/_form-input.scss
```diff
@@ -54,15 +54,16 @@
                 z-index: -1;
                 opacity: 10%;
             }
-            > .mynah-form-input {
-                background-color: transparent;
-            }
+            // > .mynah-form-input {
```
dogusata, line 57: Let's remove this one too

src/styles/components/_detailed-list.scss
```diff
@@ -42,6 +41,9 @@
         &:empty {
             display: none;
         }
+        .mynah-ui-form-item-mandatory-title {
```
dogusata, line 44: What is this doing?

src/styles/components/_detailed-list.scss
```diff
@@ -172,11 +180,10 @@
                         flex-grow: 1;
                         flex-shrink: 1;
                         color: var(--mynah-color-text-weak);
-                        font-size: var(--mynah-font-size-small);
                         display: flex;
                         overflow: hidden;
                         justify-content: flex-start;
-                        background-color: inherit;
+
```
dogusata, line 186: I couldn't see the color change to default if there is no title. like `.title + .description`.  Also, when there is no title, ellipsis should appear at the end instead of the description, so basically `dir` should be `initial`.

dogusata (changes requested):
Nice, but needs just a little more small adjustments.

src/styles/components/_detailed-list.scss
Jurredr: Yeah I know, hadn't gotten to that yet. Fixing this today

src/components/detailed-list/detailed-list.ts
Jurredr: You came up with these names, I just followed your pattern hahah ;)
But definitely agreed, will refactor.

dogusata (dismissed):

dogusata (approved):

PR #287, opened 2025-04-08
Dogusata/add switch form item

dogusata:
## Added
- `checkbox` and `switch` are added to form input types. Which accepts only `"true" | "false"` values.
- `MynahIcons.SHELL` or `"shell"` icon

## Fixed
- `tabBarButtons` are not being updated dynamically when that tab is selected. It requires to switch between tabs to let it take affect.
- `updateLastChatAnswer` doesn't update the store data if there is no messageId available. 

-------------------------
### `checkbox` and `switch` form items
```typescript
...
{
    id: 'is-online',
    type: 'checkbox',
    label: 'Yes',
    value: 'true',
    title: `Are you working online?`,
},
{
    id: 'is-monorepo',
    type: 'switch',
    icon: 'deploy',
    label: 'Yes',
    title: `Are you working in a monorepo project?`,
    tooltip: "If you're working more on monorepos, check this"
}
...
```
image

image

image

-------------------------

## Checks

## License


src/components/chat-item/chat-item-form-items.ts
```diff
@@ -106,15 +109,16 @@ export class ChatItemFormItemsWrapper {
...
+            chatOption = new Checkbox({
+              type: chatItemOption.type,
               wrapperTestId: testIds.chatItem.chatItemForm.itemToggleWrapper,
               optionTestId: testIds.chatItem.chatItemForm.itemToggleOption,
               label,
+              icon: chatItemOption.icon,
               description,
-              value,
-              options: chatItemOption.options,
+              value: value as 'true' | 'false',
```
Jurredr, line 121: Why not use a boolean?

src/components/form-items/checkbox.ts
```diff
@@ -0,0 +1,132 @@
...
+import '../../styles/components/_form-input.scss';
+
+export interface CheckboxProps {
+  type?: 'checkbox' | 'switch';
+  classNames?: string[];
+  attributes?: Record<string, string>;
+  label?: HTMLElement | ExtendedHTMLElement | string;
+  description?: ExtendedHTMLElement;
+  tooltip?: ExtendedHTMLElement;
+  value?: 'true' | 'false';
```
Jurredr, line 19: Same here, why not boolean?

src/components/icon/icons/shell.svg
Jurredr, line 1: Why is this called a shell?

src/components/form-items/checkbox.ts
dogusata: all form items should only accept string values. basic value pair for all the form items have to be just a `Record<string, string>` pair to let them pass to the consumer than backend directly.

src/components/chat-item/chat-item-form-items.ts
dogusata: all form items should only accept string values. basic value pair for all the form items have to be just a `Record<string, string>` pair to let them pass to the consumer than backend directly

Jurredr (approved):

PR #308, opened 2025-04-24
chore: move style loading to constructor

dogusata:
## Problem
Even though it is not expected, if there are two or more instances of MynahUI available in the consumer, because their style is being loaded even before they initialized, a conflict occurs between styles. 

## Solution
To avoid style conflicts, styles are being loaded when an instance is created in the constructor.

## Tests

## License


Jurredr (approved):
Small nit; remove empty constructor

PR #309, opened 2025-04-24
dogusata/move all style loading under singleton loader

dogusata:
## Problem
Styles were not being loaded through the `StyleLoader` which have the toggle to load the styles individually or skip loading them.

## Solution
Updated style loader to load individual styles dynamically. Also depending on the initialization flag.

## Tests

## License

### DIFF
--- a/src/components/card/card.ts
+++ b/src/components/card/card.ts
@@ -39,7 +39,7 @@ export class Card {
   private previousMousePosition!: { x: number; y: number };
   private mouseDownInfo!: { x: number; y: number; time: number };
   constructor (props: CardProps) {
-    StyleLoader.getInstance().load('/components/card/_card.scss');
+    StyleLoader.getInstance().load('components/card/_card.scss');
     this.props = props;
     this.render = DomBuilder.getInstance().build({
       type: 'div',
--- a/src/components/card/card.ts
+++ b/src/components/card/card.ts
@@ -3,8 +3,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import { DomBuilder, DomBuilderObject, ExtendedHTMLElement } from '../../helper/dom';
+import { StyleLoader } from '../../helper/style-loader';
 import { EngagementType } from '../../static';
-import '../../styles/components/card/_card.scss';
 
 /**
  * We'll not consider it as an engagement if the total spend time is lower than below constant and won't trigger the event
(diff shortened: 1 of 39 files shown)


src/components/card/card.ts
```diff
@@ -39,6 +39,7 @@ export class Card {
   private previousMousePosition!: { x: number; y: number };
   private mouseDownInfo!: { x: number; y: number; time: number };
   constructor (props: CardProps) {
+    StyleLoader.getInstance().load('/components/card/_card.scss');
```
Jurredr, line 42: This path starts with a '/' while others don't, is that an issue?

src/components/form-items/text-area.ts
```diff
@@ -42,6 +42,7 @@ export class TextAreaInternal extends TextAreaAbstract {
   private readyToValidate: boolean = false;
   render: ExtendedHTMLElement;
   constructor (props: TextAreaProps) {
+    StyleLoader.getInstance().load('styles/components/_form-input.scss');
```
Jurredr, line 45: `styles/` added here, while not for the other form item imports

src/components/progress.ts
```diff
@@ -25,6 +24,8 @@ export class ProgressIndicator {
   private buttonsWrapper: ChatItemButtonsWrapper;
   private props: ProgressIndicatorProps;
   constructor (props: ProgressIndicatorProps) {
+    StyleLoader.getInstance().load('components/_collapsible-content.scss');
```
Jurredr, line 27: Perhaps nice to allow multiple strings as input so they can be called through one load call?

src/styles/components/chat/_chat-wrapper.scss
```diff
@@ -163,6 +164,47 @@
...
+    &.loading {
+        > .mynah-chat-items-container {
+            padding-bottom: var(--mynah-sizing-14);
+            > .mynah-chat-items-conversation-container:last-child {
+                > .mynah-chat-item-card.mynah-chat-item-answer-stream {
+                    &:last-child {
+                        position: relative;
+                        @include chat-items-bottom-animator();
+                        > .mynah-card {
+                            min-width: 100px;
```
Jurredr, line 176: Why are we using an explicit pixel value here and not a variable of some sorts?

src/components/card/card.ts
dogusata: Doesn't matter, it skips it. I guess I forget that.

src/components/form-items/text-area.ts
dogusata: Nice catch!

src/components/progress.ts
dogusata: Each component should have one style actually. If they have more, they can be connected through the scss itself. I don't know why the collapsible-content.scss is here in progress, but didn't want to break anything right now.

Jurredr (approved):

PR #320, opened 2025-04-30
Dogusata/styling updates status changes

dogusata:
## Problem
- `header` is not possible to update partially, it requires all or none
- switch component requires restyling
- status for the header is only on the right side, it should be optional to put on the left
- spacings are not matching as the expected designs
- Refreshing content on the header causes a flickr jump

## Solution
- `header` can be updated partially, it helps keeping the user triggered collapsed/uncollapsed state of the `fileList` inside a header too. Basically every item in header can be updated partially. To update the buttons, you don't need to send the whole `header` object with its all previous values anymore.
- `switch` component is created as an individual component with its own new styles.
-  Added `position` option to `header.status` to allow consumer to put the status field to the left of the header or keep it on right.
- All spacings readjusted depending on the expected values
- Removed the reveal animation from the header blocks

## Checks

## License


docs/img/data-model/chatItems/options.png
Jurredr, line 1: There's a very slight white inset at the bottom of the selected fill it seems like, or is that just me?

Jurredr (approved):

docs/img/data-model/chatItems/options.png
dogusata: Yeap, but not on the actual one. The test ones are correct I believe.

PR #326, opened 2025-05-08
fix: updating styles in custom check card body

laileni-aws:
## Problem
- PR: https://github.com/aws/language-servers/pull/1290
- Issue if we turn status from `Allow`  to `Allowed`  in permission check in VSC Agentic flow, we should not change the header title and body so I am passing body in header and body as undefined so that previous value will be rendered. 
- This is working as expected but text in the header title just moves up if we change `Allow` to `Allowed`.
## Solution
- If we remove `margin-block-start` I was able to achieve the expected behavior.

https://github.com/user-attachments/assets/1fa567fd-db30-4d30-b70c-cbe8ee112672

## Tests

## License

### DIFF
--- a/.github/workflows/e2e_tests.yml
+++ b/.github/workflows/e2e_tests.yml
@@ -1,0 +1,57 @@
+name: E2E Tests
+on:
+  pull_request:
+    branches: [main, feature/*, fix/*]
+    types: [opened, reopened, ready_for_review, synchronize]
+jobs:
+  chromium-e2e:
+    name: Chromium E2E Tests
+    if: github.event.pull_request.draft == false
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+            fetch-depth: 20
+      - uses: actions/setup-node@v4
+        with:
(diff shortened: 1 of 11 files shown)


Jurredr:
This is not a valid solution, check out the resulting failed test snapshots to see spacings becoming inproper in e.g. (un)ordered lists. Let's discuss further on Slack.

src/styles/components/card/_card-body.scss
```diff
@@ -82,10 +82,6 @@
             word-break: normal;
             overflow-wrap: break-word;
             padding-bottom: 1px !important;
-
-            &:first-child {
```
dogusata, line 86: As i see, it causes a down shift in almost all headers. Which is unwanted and a regression. We need to do this change only for the header block i suppose. By resetting the value of margin-block-start with a stronger css specificity. And setting the value as `initial` should be enough for that.

main-spec-ts-open-mynah-ui-should-keep-the-content-inside-window-boundaries-1-snap-diff

dogusata (changes requested):

PR #331, opened 2025-05-10
fix: apply background to selection list items

awschristou:
## Problem

When open, the selection list does not have a themed background defined. It shows up in dark themes as white, and you may not see the item text.

## Solution

Apply a background to selection items.

## Tests

Verified using the sample app workbench.

image

## License


dogusata (approved):
I'm not 100% it will affect the OS based select box item bg color, but in some cases it might be supported. Nice addition.

PR #340, opened 2025-05-28
feat: add style for input border with and without focus

awschristou:
## Problem

- in themes where the default background color matches (or is similar to) the input box background, you cannot see the input box boundary
- when the input box has focus, its default color is a gradient scheme. UX direction wants this to be a standard border color

## Solution

- allow hosts to define border colors for the input box, both for when the box has focus and when it doesn't:
  - `mynah-color-text-input-border` - the color of the input box border when the box does not have focus
  - `mynah-color-text-input-border-focused` - the color of the input box border when it does have focus
- these new styles have been given default values, so that:
  - the non-focused box appearance matches the current behavior (border matches background color)
  - the focused box appearance uses the default border color (no more gradient)

## Screenshots

Before - left: unfocused input (can't see boundary), right: focused input (gradient)

image

After - left: unfocused input, right: focused input (both with sample colors)

image

## Tests
- [x] I have tested this change in the sample theme builder that is run with `npm run dev`

## License

### DIFF
--- a/ui-tests/run-browser-tests.ts
+++ b/ui-tests/run-browser-tests.ts
@@ -17,10 +17,10 @@ try {
     execSync('jest --detectOpenHandles --config=./jest.config.js', { stdio: 'inherit' });
   } else if (browser === 'chromium') {
     // Run tests for Chromium only
-    execSync('jest --updateSnapshot --detectOpenHandles --config=./jest.chromium.config.js', { stdio: 'inherit' });
+    execSync('jest --detectOpenHandles --config=./jest.chromium.config.js', { stdio: 'inherit' });
   } else if (browser === 'webkit') {
     // Run tests for Webkit only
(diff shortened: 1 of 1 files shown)


dogusata (changes requested):
Snapshots need to be updated. 

Steps:
1- Do find snapshots in both `chrome` and `webkit` folder under `./ui-tests/__test__/__image-snapshots__/`
2- Remove the snapshots which contains the prompt input field (the effective changes you made)
3- Run __(in the root folder)__ `npm run tests:e2e` to regenerate the snapshots.

awschristou:
> Snapshots need to be updated.
> 
> Steps: 1- Do find snapshots in both `chrome` and `webkit` folder under `./ui-tests/__test__/__image-snapshots__/` 2- Remove the snapshots which contains the prompt input field (the effective changes you made) 3- Run **(in the root folder)** `npm run tests:e2e` to regenerate the snapshots.

Thanks for the details. It feels like I had to update more images than necessary. I wonder if that is because I'm developing in WSL...

PR #344, opened 2025-05-30
Add padding to dropdowns

bryceitoc9:
## Problem

Dropdowns, when opened in a Chromium-based browser (most of the IDEs) don't have padding.
image

image

## Solution

Add padding, and adjust as many other elements to reduce the impact. Notably: this doubles the space between the Agentic selector and the dropdown because I'm not removing the `gap`; I'm assuming this will be good practice if we add more to this container. Happy to discuss if this looks bad (I'd probably remove or shrink the gap). Note that the padding in this case is non-negotiable since the padding on the select also drives padding on the options.

image

image

## Tests

## License

### DIFF
--- a/src/styles/components/_form-input.scss
+++ b/src/styles/components/_form-input.scss
@@ -86,7 +86,6 @@
 
             -webkit-appearance: none;
             appearance: none;
-            padding: 0;
             border: none;
             text-overflow: clip;
             outline: none;
@@ -124,6 +123,10 @@
             &.validation-error {
                 border-color: var(--mynah-color-status-error);
             }
+        }
+
+        select.mynah-form-input {
+            padding: 0px var(--mynah-sizing-2);
         }
 
         > .select-auto-width-sizer {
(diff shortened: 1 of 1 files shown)


dogusata (approved):
Please do check the tests:
image

You can find the test result assets in the relevant test workflows `upload-artifact` section. 
- Download the zip given by the link and compare the snapshots. 
- If you see the new snapshot as expected and the old one needs to be updated, you can use the images in `__received_output__` folder by removing the `-received` at the end of the file name. 
- Replace the original relevant snapshot with the one in the received folder

If you use a mac pc, you can remove the snapshots which are causing tests to fail and rerun the tests on your local to generate the updated ones too. And submit it with the PR.

dogusata (changes requested):
Please see the above comment to fix tests.

bryceitoc9:
@dogusata Screenshots updated (thanks @ctlai95 !)

laileni-aws (approved):
LGTM

PR #347, opened 2025-06-04
feat: allow mandatory form item without asterisk

ctlai95:
## Problem

When using the dropdown select component, you can use `mandatory: true` to hide the placeholder/"Please select" option from the dropdown. However doing so will also render an asterisk icon above it. We want to be able to hide the asterisk for certain cases, ie in prompt options for model selection.

image

## Solution

Only render the asterisk if mandatory is true **and** ~~the form item title is not undefined~~ new flag `hideMandatoryIcon` is not set to true 

image
image

## Tests

## License

### DIFF
--- a/docs/DATAMODEL.md
+++ b/docs/DATAMODEL.md
@@ -1073,6 +1073,7 @@ export interface ValidationPattern {
 interface BaseFormItem {
   id: string;
   mandatory?: boolean;
+  hideMandatoryIcon?: boolean;
   title?: string;
   placeholder?: string;
   value?: string;
@@ -1130,6 +1131,7 @@ export interface ListFormItem {
   type: 'list';
   id: string;
   mandatory?: boolean;
+  hideMandatoryIcon?: boolean;
   title?: string;
   description?: string;
   tooltip?: string;
@@ -2696,6 +2698,7 @@ interface ChatItemFormItem {
   id: string; // id is mandatory to understand to get the specific values for each form item when a button is clicked
   type: 'select' | 'textarea' | 'textinput' | 'numericinput' | 'stars' | 'radiogroup' | 'toggle' | 'checkbox' | 'switch' ; // type (see below for each of them)
   mandatory?: boolean; // If it is set to true, buttons in the same card with waitMandatoryFormItems set to true will wait them to be filled
+  hideMandatoryIcon?: boolean; // If it is set to true, it won't render an asterisk icon next to the form label
   title?: string; // Label of the input
   autoFocus: boolean; // focus to the input when it is created, default=> false. (Only for textual form items)
   description?: string; // The description, showing under the input field itself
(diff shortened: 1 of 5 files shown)


src/components/chat-item/chat-item-form-items.ts
```diff
@@ -49,17 +49,19 @@ export class ChatItemFormItemsWrapper {
       classNames: [ 'mynah-chat-item-form-items-container', ...(this.props.classNames ?? []) ],
       children: this.props.chatItem.formItems?.map(chatItemOption => {
         let chatOption: Select | RadioGroup | TextArea | Stars | TextInput | Checkbox | FormItemList | undefined;
-        let label: ExtendedHTMLElement | string = `${chatItemOption.mandatory === true ? '* ' : ''}${chatItemOption.title ?? ''}`;
+        let label: ExtendedHTMLElement | string = `${chatItemOption.mandatory === true && chatItemOption.title !== undefined ? '* ' : ''}${chatItemOption.title ?? ''}`;
```
awschristou, line 52: Rather than interpreting the title (eg: "if title is empty"), would it be reasonable to introduce a new optional boolean flag `hideIcon` (general, could apply elsewhere) or `hideMandatoryIcon` (more specific to use of `mandatory` property)? This property would be treated as false when not present. This way you can still allow for some future element without a title to still be decorated with the mandatory icon.

ctlai95: Yeah I was thinking of doing the boolean flag approach initially but I wanted to avoid adding a new flag. But I can see how the behavior might be confusing/unexpected. Let me give it another try

zuoyaofu:
Just curious, what is the Slack notification test? Should we need that to be approved before merging?

docs/DATAMODEL.md
```diff
@@ -2696,6 +2698,7 @@ interface ChatItemFormItem {
   id: string; // id is mandatory to understand to get the specific values for each form item when a button is clicked
   type: 'select' | 'textarea' | 'textinput' | 'numericinput' | 'stars' | 'radiogroup' | 'toggle' | 'checkbox' | 'switch' ; // type (see below for each of them)
   mandatory?: boolean; // If it is set to true, buttons in the same card with waitMandatoryFormItems set to true will wait them to be filled
+  hideMandatoryIcon?: boolean; // If it is set to true, it won't render an asterisk icon next to the form label
```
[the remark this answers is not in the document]
laileni-aws: If this wont impact other components like above ones then we should be good!

awschristou (approved):

ctlai95:
> Just curious, what is the Slack notification test? Should we need that to be approved before merging?

@zuoyaofu from what I can tell, this is just to send updates from this PR to slack, not blocking any code changes. Looks like @dogusata is working on it https://github.com/aws/mynah-ui/commit/382b828c75b70850c8ea91b304e289231f360084

PR #355, opened 2025-06-12
feat: add reroute card and add agentic notification for quickActions

Randall-Jiang:
## Problem
The current implementation of the QuickActionGenerator in the Amazon Q UI displays outdated messaging about old agents. 
## Solution
Updated the info card message in the QuickActionGenerator to add guidance about the transition to agentic.  Also, add a reroute card so when user enter /dev, /test, it would show an extra card

Screenshot 2025-06-12 at 5 07 44 AM
Screenshot 2025-06-12 at 5 08 26 AM
Screenshot 2025-06-12 at 5 08 40 AM

## Tests

## License


Randall-Jiang:
> I don't get the need for this `reroute` card, what does it have that you cannot just achieve with a regular chat item card?

agree, so I use normal chat card now

src/components/chat-item/chat-prompt-input.ts
```diff
@@ -566,12 +567,52 @@ export class ChatPromptInput {
...
+        tabId: this.props.tabId,
+        chatItem: contextHeader,
+        small: true,
+        inline: true,
+        initVisibility: true
+      });
+      children.unshift(contextHeaderCard.render);
+    }
+    // Add contextHeader from individual QuickActionCommandGroups
+    quickPickGroupList.forEach((group, index) => {
```
avi-alpert, line 598: what is contextHeader used for? I dont see an example in PR description, only for quickActionContextHeader

src/components/chat-item/chat-prompt-input.ts
```diff
@@ -566,12 +567,52 @@ export class ChatPromptInput {
...
+      if (group.contextHeader != null) {
+        const groupContextHeaderCard = new ChatItemCard({
```
avi-alpert, line 600: same with group contextHeader. I dont see an example/know the use case for this

src/helper/store.ts
```diff
@@ -22,7 +22,21 @@ const emptyDataModelObject: Required<MynahUIDataModel> = {
   cancelButtonWhenLoading: true,
   showChatAvatars: false,
   quickActionCommands: [],
+  quickActionContextHeader: {
```
avi-alpert, line 25: all additions to store.ts must contain corresponding documentation in docs/DATAMODEL.ts

src/components/chat-item/chat-item-card.ts
```diff
@@ -121,23 +120,25 @@ export class ChatItemCard {
...
-        } else {
-          this.isMoreContentExpanded = true;
-          this.render.removeClass('mynah-chat-item-collapsed');
-          this.moreContentIndicator?.update({ icon: MynahIcons.UP_OPEN });
-        }
-      },
-      testId: testIds.chatItem.moreContentIndicator
-    });
-    this.render.insertChild('beforeend', this.moreContentIndicator.render);
+    if (this.props.chatItem.autoCollapse === true || this.props.chatItem.title != null) {
```
dogusata, line 123: What is the relation with the title?

src/components/chat-item/chat-prompt-input.ts
dogusata: I couldn't get this too.

src/helper/store.ts
dogusata: Yes, it is mandatory to add the documentation, as it is indicated in the PR template, development documentation etc.

src/styles/components/chat/_chat-item-card.scss
```diff
@@ -15,6 +15,75 @@
     transform-origin: center bottom;
     gap: var(--mynah-sizing-4);
     z-index: var(--mynah-z-1);
+    &.border {
```
dogusata, line 18: We already had border on the chat-items through the Card component itself. What is the difference of that? If it was mandatory why the update didn't take place on the root component? (Card)

src/styles/components/chat/_chat-prompt-wrapper.scss
```diff
@@ -392,13 +392,21 @@
     &:not(:empty) {
         padding: var(--mynah-chat-wrapper-spacing);
         background-color: var(--mynah-card-bg);
+        border: var(--mynah-border-width) solid var(--mynah-color-border-default);
+        border-radius: var(--mynah-input-radius);
+        box-sizing: border-box;
+        margin: var(--mynah-chat-wrapper-spacing);
+        box-shadow: 0 2px 4px rgba(0, 0, 0, 0.1);
```
dogusata, line 399: Again, in the development guidelines, It is clearly indicated that you cannot put static values to anything. Use one of the matching css custom properties for the shadows or create a new one and use that please.

src/styles/components/chat/_chat-prompt-wrapper.scss
```diff
@@ -392,13 +392,21 @@
...
+        width: calc(100% - 2 * var(--mynah-chat-wrapper-spacing));
     }
 
     > .mynah-chat-item-card {
         max-width: 100%;
         width: 100%;
         > .mynah-card {
             border-bottom-left-radius: 0 !important;
+            border: none;
```
dogusata, line 408: Why we're setting cleanup values which are not being set at all? To override what comes from another usage or place? Please do update the place accordingly if you don't want to see shadow or borders instead.

dogusata (changes requested):
What we're trying with this PR is not clear. It does many changes which i'm not sure at all if they are necessary or not. 

Please first refer to the development and usage guidelines/documentations. Check already available structures for all kind of elements. 

This PR currently try to override a few things to achieve the desired look without checking what other places can be effected by them,

dogusata:
As I understood, the need is just to add  an information block on top of the context/quick action commands selector.

Please do extend the model only with what is necessary. In your case, you don't need to add context header for each context type. It is a global header for QuickPick selector. 

- Add a new field to the model
```typescript
quickPickSelectorInfo: {
  status?: Status;
  icon?: MynahIcons | MynahIconsType;
  body?: string; // [MARKDOWN]
}
```

And use it when you need it. Also create a proper component block over there. You have the icon component, you have the status color options already and a markdown parsers. Instead of using a much complex chat-item-card. You can even just use the `title-description-with-icon` component for that.

Please do investigate what is already available first as its most simple way.

src/styles/components/chat/_chat-item-card.scss
Randall-Jiang: In mynah web app, the most card does not show boarder there. That's why i add a boarder here

docs/DATAMODEL.md
```diff
@@ -551,6 +559,40 @@ const mynahUI = new MynahUI({
...
+```
+
+**Status Options:**
+- `'info'` - Blue informational styling
+- `'warning'` - Orange/yellow warning styling  
+- `'error'` - Red error styling
+- `'success'` - Green success styling
+
+<p align="center">
+  <img src="./img/data-model/tabStore/quickPickSelectorInfo.png" alt="quickPickSelectorInfo" style="max-width:500px; width:100%;border: 1px solid #e0e0e0;">
```
avi-alpert, line 591: where is this image? i dont see it in this PR

dogusata: +1

src/components/chat-item/chat-prompt-input.ts
```diff
@@ -624,7 +625,7 @@ export class ChatPromptInput {
           const searchTerm = inputValue.substring(this.quickPickTriggerIndex).match(/\S*/gi)?.[0];
           const promptRegex = new RegExp(searchTerm ?? '', 'gi');
           newQuickPickCommandGroup.commands = newQuickPickCommandGroup.commands.filter(command =>
-            command.command.match(promptRegex)
+            command.command !== '__info_card__' && command.command.match(promptRegex)
```
dogusata, line 628: You cannot hard code strings and you shouldn't do any kind of logical comparison inside mynah-ui. MynahUI doesn't know that what info card is which comes through a command. You cannot check commands. That check and the flow should only be done through the consumer app.

src/components/chat-item/chat-prompt-input.ts
```diff
@@ -640,7 +641,20 @@ export class ChatPromptInput {
   };
 
   private readonly getQuickPickItemGroups = (quickPickGroupList: QuickActionCommandGroup[]): ExtendedHTMLElement => {
-    const detailedListItemsGroup = convertQuickActionCommandGroupsToDetailedListGroups(quickPickGroupList);
+    // Extract info card data and filter out special commands
+    const filteredQuickPickGroupList = quickPickGroupList.map(group => ({
+      ...group,
+      // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+      commands: group.commands?.filter(cmd => {
+        if (cmd.command === '__info_card__') {
```
dogusata, line 649: Same here, you cannot do comparison for text. What if we have another in the future, and than another. MynahUI is not the place that checks statments in a logical manner.

src/helper/store.ts
```diff
@@ -22,6 +22,7 @@ const emptyDataModelObject: Required<MynahUIDataModel> = {
   cancelButtonWhenLoading: true,
   showChatAvatars: false,
   quickActionCommands: [],
+  quickPickSelectorInfo: { status: undefined, icon: undefined, body: undefined },
```
dogusata, line 25: why not just `{}`

src/static.ts
```diff
@@ -88,6 +88,14 @@ export interface MynahUIDataModel {
   */
   quickActionCommands?: QuickActionCommandGroup[];
   /**
+  * Information block to show on top of quick pick selector
+  */
+  quickPickSelectorInfo?: {
+    status?: Status;
+    icon?: MynahIcons | MynahIconsType;
+    body?: string; // [MARKDOWN]
```
dogusata, line 96: You don't need to keep the comment.

src/styles/_variables.scss
```diff
@@ -119,6 +119,7 @@
     --mynah-shadow-button: none;
     --mynah-shadow-card: none;
     --mynah-shadow-card-hover: 0 40px 25px -15px rgba(0, 0, 0, 0.35);
+    --mynah-shadow-card-border: 0 2px 8px rgba(0, 0, 0, 0.08);
```
dogusata, line 122: It is gonna work with all themes? Both dark and light, not asking for a change but asking for confirmation.

src/styles/components/chat/_chat-item-card.scss
dogusata: Ok, but the card component already has the `border` functionality. This is still duplicating the available styling. Instead you could just give to the related card `border: true`. I still don't get why we're adding another border styling?

src/styles/components/chat/_chat-item-card.scss
```diff
@@ -15,6 +15,75 @@
...
+        margin: var(--mynah-sizing-4) 0;
+        > .mynah-card {
```
dogusata, line 20: Same here, why we have another mynah-card duplication here? What we're gonna do, every time we need a new card are we gonna duplicate the styling instead of adding more options to the base card component?

dogusata (changes requested):

PR #368, opened 2025-06-18
fix: select dropdown icon is not clickable

ctlai95:
## Problem

Sometimes dropdown arrow icon is not clickable.

image

## Solution

The issue is that the select element width is getting overridden to `auto`. This change will ensure it uses `100%` instead.

## Tests

## License


dogusata (approved):

laileni-aws (approved):
LGTM

PR #374, opened 2025-06-23
allow both capitalized and all lowercases command name used for image…

yzhangok:
… context

## Problem
For the context provided by language-server, all commands are capitalized, 
Screenshot 2025-06-23 at 3 32 57 PM

and in the context provided in `example`, commands name are all lower cases. 
Screenshot 2025-06-23 at 3 33 34 PM

## Solution
To make the components able to handle image context provided by both cases(from example's config & language-server context provider), add logic of to lower cases the command name. 

## Tests

## License

### DIFF
--- a/example/src/main.ts
+++ b/example/src/main.ts
@@ -1079,7 +1079,7 @@ here to see if it gets cut off properly as expected, with an ellipsis through cs
         },
 
         onContextSelected(contextItem, tabId) {
-            if (contextItem.command === 'image') {
+            if (contextItem.command.toLowerCase() === 'image') {
                 const fileInput = document.createElement('input');
                 fileInput.type = 'file';
                 fileInput.accept = 'image/*';
--- a/src/components/chat-item/chat-wrapper.ts
+++ b/src/components/chat-item/chat-wrapper.ts
@@ -506,7 +506,7 @@ export class ChatWrapper {
   private hasImageContextCommand (): boolean {
     const contextCommands = MynahUITabsStore.getInstance().getTabDataStore(this.props.tabId).getValue('contextCommands') as QuickActionCommandGroup[] | undefined;
     return !((contextCommands?.some(group =>
-      group.commands.some((cmd: QuickActionCommand) => cmd.command === 'image')
+      group.commands.some((cmd: QuickActionCommand) => cmd.command.toLowerCase() === 'image')
     )) === false);
   }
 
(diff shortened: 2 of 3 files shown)


example/src/main.ts
```diff
@@ -1079,7 +1079,7 @@ here to see if it gets cut off properly as expected, with an ellipsis through cs
         },
 
         onContextSelected(contextItem, tabId) {
-            if (contextItem.command === 'image') {
+            if (contextItem.command.toLowerCase() === 'image') {
```
laileni-aws, line 1082: How does this lowerCase other Commands apart from image?

ctlai95 (approved):

example/src/main.ts
yzhangok: Other commands don't need lower case check. As this is in `example`, a same logic is implemented in chat-client for the 'real' case. In chat-client, image context is routed to a special handling other than other context.

laileni-aws (approved):

src/components/chat-item/chat-wrapper.ts
```diff
@@ -506,7 +506,7 @@ export class ChatWrapper {
   private hasImageContextCommand (): boolean {
     const contextCommands = MynahUITabsStore.getInstance().getTabDataStore(this.props.tabId).getValue('contextCommands') as QuickActionCommandGroup[] | undefined;
     return !((contextCommands?.some(group =>
-      group.commands.some((cmd: QuickActionCommand) => cmd.command === 'image')
+      group.commands.some((cmd: QuickActionCommand) => cmd.command.toLowerCase() === 'image')
```
dogusata, line 509: Again, please do not use static check statments. MynahUI doesn't know what 'image' command is. It shouldn't know about that. The placement of this check is not inside MynahUI. You can do it only while representing the text, even without effecting the filtering. Or you can update the filter to lower case the text and than match.

dogusata (changes requested):
**MynahUI is not/shouldn't be responsible to check any single information coming from the data itself.** 

Do not use data dependent check statements, do not add logical steps depending on the incoming data content and do not use static checks in any single place. 

Assume that in the upcoming weeks we'll have `document` and the week after that `configuration`. If we're planning to add them to here again or extend the if statment with multiple ORs, this is wrong. 

MynahUI is not responsible to check any single information coming through data and it shouldn't rely on the incoming data content for any checks or logic.

PR #376, opened 2025-06-24
Feat: Add a template for a bordered and padding card

Randall-Jiang:
## Problem
When user send `/` quickAction, we should return a reroute warning card
## Solution
I add a border field in ChatItem interface and modify chat-item-card scss file. I tried to use global border variable in _card.scss however it does not work. Because we removed all border for chat-item-card for agentic Chat. Since the border filed is a new added, so it won't affect any existing component. (e2e test 100% pass). Border would only apply when `border: true`.

Screenshot 2025-06-23 at 10 27 08 PM

## Tests

## License


dogusata (dismissed):

example/src/config.ts
```diff
@@ -215,6 +215,12 @@ export const QuickActionCommands: QuickActionCommandGroup[] = [
                 description:
                     'You can set the position of the followups too. By simply setting the type of the ChatItem.',
             },
+            {
+                command: Commands.REROUTE_CARDS,
```
ctlai95, line 219: Isn't it too specific to call this "reroute cards"? It's just a normal card with a border?

src/components/chat-item/chat-item-card.ts
```diff
@@ -112,8 +112,20 @@ export class ChatItemCard {
...
       children: this.initialSpinner ?? [],
       background: this.props.inline !== true && this.props.chatItem.type !== ChatItemType.DIRECTIVE && !(this.props.chatItem.fullWidth !== true && (this.props.chatItem.type === ChatItemType.ANSWER || this.props.chatItem.type === ChatItemType.ANSWER_STREAM)),
-      border: this.props.inline !== true && this.props.chatItem.type !== ChatItemType.DIRECTIVE && !(this.props.chatItem.fullWidth !== true && (this.props.chatItem.type === ChatItemType.ANSWER || this.props.chatItem.type === ChatItemType.ANSWER_STREAM)),
-      padding: this.props.inline === true || this.props.chatItem.padding === false || (this.props.chatItem.fullWidth !== true && (this.props.chatItem.type === ChatItemType.ANSWER || this.props.chatItem.type === ChatItemType.ANSWER_STREAM)) ? 'none' : undefined,
+      border: (this.props.inline !== true &&
+        this.props.chatItem.type !== ChatItemType.DIRECTIVE &&
+        !(this.props.chatItem.fullWidth !== true &&
+          (this.props.chatItem.type === ChatItemType.ANSWER ||
+           this.props.chatItem.type === ChatItemType.ANSWER_STREAM))) ||
+        // Auto border for warning/info headers
```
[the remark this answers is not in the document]
Randall-Jiang: if we pass border = true then all cards would have borders. But for agentic chat we are canceling all the borders rigth?

example/src/config.ts
Randall-Jiang: Yes but we want to show this card as an example in mynah webapp as an example

ctlai95: Sure but I don't think we need to create a "reroute cards" section for this. That is something very specific to agentic chat and should be decoupled from this library.

For example "Sticky card" could be called "Model unavailable card" but that would be too specific here

src/components/chat-item/chat-item-card.ts
```diff
@@ -112,8 +112,18 @@ export class ChatItemCard {
...
+        (this.props.chatItem.border === true),
+      padding: ((this.props.inline === true || this.props.chatItem.padding === false || (this.props.chatItem.fullWidth !== true && (this.props.chatItem.type === ChatItemType.ANSWER || this.props.chatItem.type === ChatItemType.ANSWER_STREAM))) &&
+        // Exception: warning/info headers should have padding
+        !(this.props.chatItem.header?.padding === true))
```
avi-alpert, line 124: this boolean logic is very confusing, you;'re checking if `this.props.chatItem.padding === false` but also `!(this.props.chatItem.header?.padding === true))`

example/src/config.ts
laileni-aws: Yes, I agree, this repo should be platform Agnostic, 
@Randall-Jiang Lets rename this to any  `Sample` generic text.

example/src/samples/sample-data.ts
```diff
@@ -447,30 +447,35 @@ Transform your java project from an old version to a new one.
...
-                description: 'Generate code across files with a task description.',
-                placeholder: 'Type your question',
+                description: 'Generate code to make a change in your project',
+                placeholder: 'Describe your task or issue in as much detail as possible',
             },
             {
-                command: '/write',
-                icon: MynahIcons.BUG,
-                description: 'Automatically write code and commit it',
+                command: '/doc',
```
laileni-aws, line 458: This can be write thats fine

example/src/samples/sample-data.ts
```diff
@@ -1286,6 +1296,28 @@ export const exampleInformationCard = (
     };
 };
 
+export const exampleRouteCard = (): ChatItem => {
```
laileni-aws, line 1299: We can make this even very generic!

laileni-aws (approved):

PR #377, opened 2025-06-25
fix: radiogroup toggle will always stack vertically

chungjac:
## Problem
- Radiogroup toggle items are stacked horizontally if window is expanded enough
- When radiogroup toggle item is not selected, it has a circle inside. This has confused customers as it looks like it is selected even though it is not. 

https://github.com/user-attachments/assets/dcaf3fc8-f0f5-4068-999a-ea467d38c35a

## Solution
- Radiogroup toggle items will always stack vertically, regardless of window size
- When radiogroup toggle item is not selected, it will be blank inside the circle
- Increase spacing between the toggle and the text 

https://github.com/user-attachments/assets/2f17da0e-3141-45a6-a0e7-60502e9fbea9

## Tests

## License


dogusata (changes requested):
Very nice PR! Thanks.

But,
It will require the tests to be updated as well. Also, screenshots used in documentation. Please do update them accordingly too then it is good to merge.

chungjac:
> Very nice PR! Thanks.
> 
> But, It will require the tests to be updated as well. Also, screenshots used in documentation. Please do update them accordingly too then it is good to merge.

Thank you, will update the tests and documentation today

laileni-aws (approved):

PR #387, opened 2025-07-02
Add space and move the cursor after drag and drop to empty prompt, ch…

yzhangok:
…ange image icon and improve overlay removal's latency

## Problem
This is to fix the issue found during bug bash
1) The image icon looks different from other icon
2) The latency of the removal of drag overlay is high
3) When prompt input is empty, context added by drag&drop is not separated by empty space, and the cursor is not moving to the end of the input
4) The images added by drag and drop sometimes are added to pinned context

## Solution
1) Use the right icon

Screenshot 2025-07-01 at 10 30 08 PM

2) Force repaint before dispatch the event

3) The `insertContextToPosition` is not able to handle the case when cursor position is 0, this is only happen in drag&drop, since customer does not need to type/click to trigger add context.

4) This is reproducable when first click on `@Pin Context`, then does not select anything. Since the `topBarTitleClick` is still true in this case, the context added through drag&drop will be added to pinned context. Dispatch the event `RESET_TOP_BAR_CLICKED` (Renamed from `CONTEXT_INSERTED`) to reset the trigger source before add the context.

## Tests

## License


laileni-aws (dismissed):
LGTM but this image looks odd to me in `src/components/icon/icons/image.svg` but looks good in PR description though.

One more observation: Icon is little up compared to Icon title in below screenshot.
Screenshot 2025-07-01 at 10 30 08 PM

src/static.ts
```diff
@@ -256,7 +256,7 @@ export enum MynahEventNames {
   TOP_BAR_BUTTON_CLICK = 'promptInputTopBarButtonClick',
   CONTEXT_PINNED = 'contextPinned',
   FILES_DROPPED = 'filesDropped',
-  CONTEXT_INSERTED = 'contextInserted'
+  RESET_TOP_BAR_CLICKED = 'resetToBarClicked'
```
avi-alpert, line 259:
```suggestion
  RESET_TOP_BAR_CLICKED = 'resetTopBarClicked'
```

dogusata (changes requested):
I'm not fan of strictly defining the whole drag drop functionality just for images. This will cause duplication of code whenever we need different file types to be handled. However, you can do note it as a future `TODO`.

However, please avoid using hard coded strings for check statements. It is pretty open to mistakes. And not future proof. Please move the `image` string comparison to type. in the `QuickActionCommand` type definition `command` can be `string | ImageCommandTypes` which you can define. Then the consumer has to follow it.

src/components/chat-item/chat-wrapper.ts
```diff
@@ -149,6 +154,13 @@ export class ChatWrapper {
       }
     });
 
+    MynahUITabsStore.getInstance().addListenerToDataStore(props.tabId, 'contextCommands', (contextCommands: QuickActionCommandGroup[]) => {
+      // Feature flag for image context command
+      this.imageContextFeatureEnabled = contextCommands?.some(group =>
+        group.commands.some((cmd: QuickActionCommand) => cmd.command.toLowerCase() === 'image')
```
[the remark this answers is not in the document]
yzhangok: This is the feature flags, will be removed after image context is launched in all IDEs.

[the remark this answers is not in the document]
yzhangok: The IDE controls image-related functionality through the imageContextEnabled feature flag. When this flag is set to true, the language server adds the Image option to the available context types.

Users can add images to the context in Mynah UI through three methods:

1. Using the context command menu
2. Typing the @image: command
3. Dragging and dropping images
To maintain consistency, we've implemented a centralized feature flag that controls the visibility of all three image-adding methods. This ensures that image functionality is either entirely available or unavailable across for an IDE.

src/components/chat-item/chat-wrapper.ts
```diff
@@ -179,6 +191,21 @@ export class ChatWrapper {
       this.promptInputElement = this.promptInput.render;
     }
 
+    // Always-present drag overlays (hidden by default, shown by style)
+    this.dragOverlayContent = DomBuilder.getInstance().build({
+      type: 'div',
+      classNames: [ 'mynah-drag-overlay-content' ],
+      children: [
+        new Icon({ icon: MynahIcons.IMAGE }).render,
+        { type: 'span', children: [ 'Add image to context' ] }
```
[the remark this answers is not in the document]
yzhangok: Fixed

src/static.ts
```diff
@@ -789,6 +790,7 @@ export interface ConfigOptions {
   codeInsertToCursorEnabled?: boolean;
   codeCopyToClipboardEnabled?: boolean;
   test?: boolean;
+  dragOverlayIcon?: MynahIcons | MynahIconsType | CustomIcon;
```
avi-alpert, line 793: TODO: Please add documentation to CONFIG.md for these new fields

yzhangok: Will send followup pr to fix this

src/components/chat-item/chat-wrapper.ts
```diff
@@ -66,6 +71,26 @@ export class ChatWrapper {
       type: 'div',
       classNames: [ 'mynah-chat-wrapper-header-spacer' ]
     });
+
+    /*
+    The IDE controls image-related functionality through the imageContextEnabled feature flag.
```
avi-alpert, line 76: This comment is not true. You did not add an imageContextEnabled feature flag that the IDE can control. This PR sets imageContextEnabled based on whether an `image` command is included in the contextCommands

Please either fix the comment or preferably, fix the functionality and have it as a feature flag set in config

src/components/chat-item/chat-wrapper.ts
```diff
@@ -283,7 +295,30 @@ export class ChatWrapper {
...
+            .mynah-drag-overlay-content {
+              position: absolute;
+              top: 50%;
+              left: 50%;
+              transform: translate(-50%, -50%);
+              display: flex;
+              flex-direction: column;
+              align-items: center;
+              z-index: 2;
+              font-size: 1.2em;
```
avi-alpert, line 318: do not use hardcoded font size, use font size variable

src/components/chat-item/chat-wrapper.ts
```diff
@@ -283,7 +295,30 @@ export class ChatWrapper {
...
+              color: #333;
```
avi-alpert, line 319: another hardcoded color

src/static.ts
avi-alpert: yeah thats fine

avi-alpert (approved):
I will approve to unblock launch - however please refer back to this PR's comments to be addressed in a follow up

src/components/chat-item/chat-wrapper.ts
laileni-aws: +1

dogusata:
As Avi mentioned, please do take it as a first task after release and address all given comments above.

PR #389, opened 2025-07-08
fix: move pinContextHint to config texts

avi-alpert:
## Problem
The `Pin context with ⌥ Enter` hint text was hardcoded which violates the guidelines of having all strings in config texts. The consumer needs to be able to override this string since Windows machines use `Alt Enter` 

## Solution
Move hardcoded pin context hint string to config texts so consumers can override string.

## Tests

## License


dogusata:
Only thing, it might be better to use the code for the OPTION character, in case there will be some different type of encodings.

dogusata (approved):

PR #393, opened 2025-07-11
make the escape prompt logic able to handle context command name with…

yzhangok:
… special character and space

## Problem
The context is not formatted correctly if there's an empty space 

image

## Solution
Instead of search for `@`, use the context list to construct regex used to find in the prompt text.

Screenshot 2025-07-10 at 7 49 41 PM

## Tests

## License

### DIFF
--- a/src/components/chat-item/chat-prompt-input.ts
+++ b/src/components/chat-item/chat-prompt-input.ts
@@ -848,11 +848,19 @@ export class ChatPromptInput {
         : currentInputValue + (attachmentContent ?? '');
       const context: QuickActionCommand[] = this.promptTextInput.getUsedContext();
 
-      const escapedPrompt = escapeHTML(promptText.replace(/@\S*/gi, (match) => {
-        // Unnecessary spaces will be removed by HTML rendering,
-        // it is safe to add a start space to avoid rendering problems for bold text
-        return ` **${match}**`;
-      }));
+      let escapedPrompt = escapeHTML(promptText);
+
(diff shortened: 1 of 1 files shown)


src/components/chat-item/chat-prompt-input.ts
```diff
@@ -848,11 +848,19 @@ export class ChatPromptInput {
...
       const context: QuickActionCommand[] = this.promptTextInput.getUsedContext();
 
-      const escapedPrompt = escapeHTML(promptText.replace(/@\S*/gi, (match) => {
-        // Unnecessary spaces will be removed by HTML rendering,
-        // it is safe to add a start space to avoid rendering problems for bold text
-        return ` **${match}**`;
-      }));
+      let escapedPrompt = escapeHTML(promptText);
+
+      if (Array.isArray(context)) {
```
laileni-aws, line 853: nit: I guess we can simplify this

```
let escapedPrompt = escapeHTML(promptText);
context?.forEach(cmd => {
  if (cmd.command) {
    // Escape special regex characters in the command
    const escapedCmd = cmd.command.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
    // Replace all occurrences of @command with **@command**
    escapedPrompt = escapedPrompt.replace(
      new RegExp(`@${escapedCmd}`, 'g'), 
      ` **@${cmd.command}**`
    );
  }
});
```

laileni-aws (approved):

laileni-aws (approved):

PR #396, opened 2025-07-15
feat: call onSearchShortcut on ctrl/cmd + f

avi-alpert:
### Problem
Currently, there is no way to intercept and handle the standard search shortcut (Cmd+F/Ctrl+F) within Mynah UI. Users want to be able to open chat history search with a keyboard shortcut.

### Solution
Changes include:

- Added `enableSearchKeyboardShortcut` to config so consumers can opt-in to enable the keyboard shortcut 
- Added a new `onSearchShortcut` event handler that fires on Cmd+F or Ctrl+F
- Prevents the default search behavior
- Provides the current tab ID and event ID to the consumer for custom implementation
- Added documentation in `PROPERTIES.md` and `CONFIG.md`
- Implemented the event handler in the main MynahUI class

## Tests
- [x] I have tested this change on JetBrains

## License

### DIFF
--- a/docs/PROPERTIES.md
+++ b/docs/PROPERTIES.md
@@ -1296,3 +1296,23 @@ onFilesDropped: (tabId, files, insertPosition) => {
 ```
 
 ---
+
+### `onSearchShortcut`
+
+This event will be fired when the user presses Command+F (Mac) or Ctrl+F (Windows/Linux). It passes the `tabId` of the current tab and the `eventId` for tracking user intent. This allows the consumer to implement custom search functionality when the standard browser search shortcut is pressed.
+
+```typescript
+...
+onSearchShortcut?: (
+    tabId: string,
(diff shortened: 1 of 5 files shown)


docs/PROPERTIES.md
```diff
@@ -1296,3 +1296,23 @@ onFilesDropped: (tabId, files, insertPosition) => {
...
 
 ---
+
+### `onSearchShortcut`
+
+This event will be fired when the user presses Command+F (Mac) or Ctrl+F (Windows/Linux). It passes the `tabId` of the current tab and the `eventId` for tracking user intent. This allows the consumer to implement custom search functionality when the standard browser search shortcut is pressed.
+
+```typescript
+...
+onSearchShortcut?: (
```
laileni-aws, line 1306: Nice

PR #397, opened 2025-07-15
add public setDragOverlayVisible and resetTopBarClicked methods

yzhangok:
## Problem
Fix the bug
1) Dragging an image sometimes adds it to pinned context instead of in the prompt 
2) Drag overlay is not rendered 

## Solution
* Add two public method that can be used by consumer
 1. Show/Hide drag overlay
 2. Reset the topBarClicked, this is the variable controls whether context is added to pinned context or prompt

## Tests
- [ x ] I have tested this change on JetBrains

## License


yzhangok:
Updated the problem statement & solution

avi-alpert (approved):

PR #398, opened 2025-07-16
fix: UI Improvements for Non-Chat Tabs and Icon Wrapper Optimization

laileni-aws:
## Problem
- When no icon is provided in `tabHeaderDetails`, the `mynah-ui-title-description-icon-wrapper` class adds unwanted left padding. 
image

## Solution
- Modified `title-description-with-icon.ts` to conditionally apply icon wrapper styles.
- Only applies the `mynah-ui-title-description-icon-wrapper` class when an icon is provided
- Improves layout consistency when no icon is present
- The changes have been tested in the example application and work as expected.
image

## Tests

## License

### DIFF
--- a/src/components/title-description-with-icon.ts
+++ b/src/components/title-description-with-icon.ts
@@ -24,7 +24,11 @@ export class TitleDescriptionWithIcon {
     this.render = DomBuilder.getInstance().build({
       type: 'div',
       testId: props.testId,
-      classNames: [ 'mynah-ui-title-description-icon-wrapper', ...(this.props.classNames ?? []) ],
+      // Apply icon wrapper styles only if icon is provided
+      classNames: [
+        ...(this.props.icon !== undefined ? [ 'mynah-ui-title-description-icon-wrapper' ] : []),
+        ...(this.props.classNames ?? [])
+      ],
       children: [
         ...(this.props.icon !== undefined
           ? [ {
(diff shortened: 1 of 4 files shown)


src/components/title-description-with-icon.ts
```diff
@@ -24,7 +24,11 @@ export class TitleDescriptionWithIcon {
     this.render = DomBuilder.getInstance().build({
       type: 'div',
       testId: props.testId,
-      classNames: [ 'mynah-ui-title-description-icon-wrapper', ...(this.props.classNames ?? []) ],
+      // Apply icon wrapper styles only if icon is provided
+      classNames: [
+        ...(this.props.icon !== undefined ? [ 'mynah-ui-title-description-icon-wrapper' ] : []),
```
avi-alpert, line 29: this could be a breaking change since this class is also used in sheet.scss and detailed-list.scss

laileni-aws: Ideally If icon does not exists we should not add icon wrapper styles. Do you know where exactly these two styles are being used ? for which component so that I can test for these component too!

laileni-aws: Tested in sheet.css in MCP, no regression here and I am trying to fix for `detailed-list.scss`.
image

awschristou (approved):

PR #402, opened 2025-07-16
Fix incorrect layering bug with status icon in WebKit

shruti0085:
## Description

The status icon showed in chat items caused crashes for extensions for IDEs such as Eclipse which uses WebKit based browsers on Mac. Error : `WebView::GraphicsContextCG::endTransparentLayer()` would be thrown because the `mask-size: 100%,60%` is meant to be used with images having multiple layers. For our use case, we only seem to use single layers and WebKit's CoreGraphics renderer has strict validation for mask operations. The invalid comma syntax was causing an assertion failure in CoreGraphics, which triggered the crash. To fix this, since we only have a single layer we now specify `mask-size: 100% 60%` as width and height respectively. 
before
Screenshot 2025-07-16 at 4 07 04 PM

after
Screenshot 2025-07-16 at 4 05 28 PM

Diff 
Screenshot 2025-07-16 at 4 35 32 PM

## Tests

## License


laileni-aws (approved):
Awesome!
Can you share the screenshots! Please 

Thanks

avi-alpert (approved):

PR #403, opened 2025-07-16
feat: add option to disable dropdown form item

dungdong-aws:
## Problem

Cannot disable dropdown/select form item. We to support this for more use cases

## Solution

Add `disabled` prop to `DropdownFormItem` to disable it in `ChatItemFormItemsWrapper`

## Screenshots

Screenshot 2025-07-16 at 4 05 52 PM

VSC Screenshot:

Screenshot 2025-07-16 at 4 04 08 PM

## Tests

## License

### DIFF
--- a/src/static.ts
+++ b/src/static.ts
@@ -514,6 +514,7 @@ type DropdownFormItem = BaseFormItem & {
     value: string;
     label: string;
   }>;
+  disabled?: boolean;
 };
 
 type Stars = BaseFormItem & {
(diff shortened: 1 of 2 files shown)


src/static.ts
```diff
@@ -514,6 +514,7 @@ type DropdownFormItem = BaseFormItem & {
     value: string;
     label: string;
   }>;
+  disabled?: boolean;
```
abhraina-aws, line 517: Heard that this is current production behavior, but is it safe to do this? Allow by default to write in the workspace?

laileni-aws: Do we have documentation for this ? If yes can you update it there?

dungdong-aws: Updated it! @laileni-aws

ashishrp-aws (approved):

laileni-aws (approved):

PR #405, opened 2025-07-21
feat: add description to options in detailed list and tooltip for select option

dungdong-aws:
## Problem

New UX/UI requirement to increase the gap in for `filterOptions` in `detailed-list` component. Also add a small description to a option & bold the title of a FormItem

Also, we also need a tooltip for the formItem itself, not just the Wrapper object. So if we point to an option, the tooltip should be right above it, not the Wrapper.

## Solution

For adding description: add another option below the current option, then disable it

For adding tooltip: define the container object for the dropdown, add event handlers to show/hide tooltip within that object. 

For boldTitle: add a boolean prop to check then make it `bold` on init

### Tooltip: 

Screenshot 2025-07-21 at 10 07 07 AM

### Description:

Screenshot 2025-07-21 at 10 07 26 AM

### Gap & bold title
Screenshot 2025-07-21 at 10 07 44 AM

### VSC Screenshots

Screenshot 2025-07-21 at 10 09 48 AM

Screenshot 2025-07-21 at 10 10 13 AM

## Tests

## License


avi-alpert:
Curious how this looks on JB, I know they've add issues with these dropdowns rendering properly

dungdong-aws:
> Curious how this looks on JB, I know they've add issues with these dropdowns rendering properly

Hi @avi-alpert, this is the screenshot on JB: 

Screenshot 2025-07-21 at 1 22 21 PM

avi-alpert (approved):

src/components/chat-item/chat-item-form-items.ts
```diff
@@ -49,7 +49,12 @@ export class ChatItemFormItemsWrapper {
       classNames: [ 'mynah-chat-item-form-items-container', ...(this.props.classNames ?? []) ],
       children: this.props.chatItem.formItems?.map(chatItemOption => {
         let chatOption: Select | RadioGroup | TextArea | Stars | TextInput | Checkbox | FormItemList | undefined;
-        let label: ExtendedHTMLElement | string = `${chatItemOption.mandatory === true && chatItemOption.hideMandatoryIcon !== true ? '* ' : ''}${chatItemOption.title ?? ''}`;
+        let label: ExtendedHTMLElement | string = chatItemOption.boldTitle === true
+          ? DomBuilder.getInstance().build({
+            type: 'strong',
+            children: [ `${chatItemOption.mandatory === true && chatItemOption.hideMandatoryIcon !== true ? '* ' : ''}${chatItemOption.title ?? ''}` ]
```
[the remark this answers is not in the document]
dungdong-aws: Ah I see, let me clean it. Thanks for the catch!

src/static.ts
```diff
@@ -513,8 +514,10 @@ type DropdownFormItem = BaseFormItem & {
   options?: Array<{
     value: string;
     label: string;
+    description?: string;
   }>;
   disabled?: boolean;
+  selectTooltip?: string;
```
avi-alpert, line 520: I think it would be better to give it a more generic name like `tooltip`. In the documentation you can say that initially we are only supporting it for select, however in the future we might want other fields to have tooltips as well

```suggestion
  tooltip?: string;
```

src/components/chat-item/chat-item-form-items.ts
```diff
@@ -48,8 +48,14 @@ export class ChatItemFormItemsWrapper {
       testId: testIds.chatItem.chatItemForm.wrapper,
       classNames: [ 'mynah-chat-item-form-items-container', ...(this.props.classNames ?? []) ],
       children: this.props.chatItem.formItems?.map(chatItemOption => {
+        const title = `${chatItemOption.mandatory === true && chatItemOption.hideMandatoryIcon !== true ? '* ' : ''}${chatItemOption.title ?? ''}`;
         let chatOption: Select | RadioGroup | TextArea | Stars | TextInput | Checkbox | FormItemList | undefined;
-        let label: ExtendedHTMLElement | string = `${chatItemOption.mandatory === true && chatItemOption.hideMandatoryIcon !== true ? '* ' : ''}${chatItemOption.title ?? ''}`;
+        let label: ExtendedHTMLElement | string = chatItemOption.boldTitle === true
+          ? DomBuilder.getInstance().build({
+            type: 'strong',
```
avi-alpert, line 55: This is a valid approach but I don't think `<strong>` is the best way to go here. 

I would recommend to solve this problem with css (`font-weight: bold`). You can add a class (call it something like `'mynah-ui-form-item-bold-label'`) to `src/styles/components/_form-input.scss` under `.mynah-form-input-label`

Our repo's best practice is to use CSS for styling when possible, which helps maintain consistency across our project. (as mentioned in DEVELOPER.md)

dungdong-aws: Updated it!

avi-alpert (approved):

src/static.ts
dungdong-aws: Yea, its reference is the wrapper --> always show on top of title

PR #411, opened 2025-07-23
fix: add log by using Log() function

yzhangok:
## Solution

Use the Console on the right side of the MynahUI example page for logging.

Screenshot 2025-07-23 at 3 34 00 PM

## Tests

## License


chungjac (dismissed):

example/src/main.ts
```diff
@@ -1112,6 +1113,7 @@ here to see if it gets cut off properly as expected, with an ellipsis through cs
                         };
 
                         mynahUI.addCustomContextToPrompt(tabId, [contextItem])
+                        Log(`Image context added: <b>${contextItem.command}</b>`);
```
avi-alpert, line 1116: This is the same log message as the one above, could we differentiate it? So it's clear in the log what method was used

PR #414, opened 2025-07-24
fix: decrease dropdown arrow spacing and always have border around prompt input box

chungjac:
## Problem
- dropdown arrow spacing has too much spacing
- font size for dropdown options is too large
- we don't have a border if prompt input box is not in focus

## Solution
- decrease spacing for dropdown arrow 
- decrease dropdown font size from `mynah-font-size-small` to `mynah-font-size-xsmall`
- always have border around prompt input box even when not in focus

#### Old mynahui:
old

#### New mynahui:
new

#### Old and new VSC:
old vsc

new vsc

#### Out of focus border:
out of focus

#### In focus border:
in focus

#### Now the border always shows (regardless of in focus or out of focus):
https://github.com/user-attachments/assets/6795de13-505f-48b0-86ab-f2b646481ba1

## Tests

## License

### DIFF
--- a/src/styles/components/chat/_chat-prompt-wrapper.scss
+++ b/src/styles/components/chat/_chat-prompt-wrapper.scss
@@ -250,13 +250,14 @@
                                 }
                                 > .mynah-form-input-container.no-border {
                                     > .select-auto-width-sizer {
+                                        font-size: var(--mynah-font-size-xsmall);
                                         padding-right: calc(var(--mynah-sizing-half) + 1em);
                                     }
                                     > select {
                                         width: auto;
                                         min-width: auto;
                                         color: var(--mynah-color-text-default);
-                                        font-size: var(--mynah-font-size-small);
+                                        font-size: var(--mynah-font-size-xsmall);
                                         & + i {
                                             font-size: var(--mynah-font-size-xsmall);
                                         }
@@ -392,7 +393,7 @@
             pointer-events: none;
         }
         &::before {
-            border: var(--mynah-border-width) solid var(--mynah-color-text-input-border);
+            border: var(--mynah-border-width) solid var(--mynah-color-text-input-border-focused);
         }
     }
 


laileni-aws (approved):

src/styles/components/chat/_chat-prompt-wrapper.scss
```diff
@@ -250,13 +250,14 @@
...
                                     > .select-auto-width-sizer {
+                                        font-size: var(--mynah-font-size-xsmall);
                                         padding-right: calc(var(--mynah-sizing-half) + 1em);
                                     }
                                     > select {
                                         width: auto;
                                         min-width: auto;
                                         color: var(--mynah-color-text-default);
-                                        font-size: var(--mynah-font-size-small);
+                                        font-size: var(--mynah-font-size-xsmall);
```
avi-alpert, line 260: I see the styling changes include 2 font size changes and a border color change. Which change corresponds with `decrease spacing for dropdown arrow`? Did that happen just as a side effect of the font size reduction?

chungjac: This corresponds to `decrease spacing for dropdown arrow`:
```
  > .select-auto-width-sizer {
      font-size: var(--mynah-font-size-xsmall);
      padding-right: calc(var(--mynah-sizing-half) + 1em);
  }
```


This corresponds to `decrease dropdown font size `:
```
  > select {
      width: auto;
      min-width: auto;
      color: var(--mynah-color-text-default);
      font-size: var(--mynah-font-size-xsmall);
      & + i {
          font-size: var(--mynah-font-size-xsmall);
      }
  }
```


My understanding:
- `.select-auto-width-sizer` is an invisible element that contains the same text as the selected option. When we change the `select` font size, the visible text becomes smaller. However the sizer still uses its own font size to calculate width, causing the spacing discrepancy

avi-alpert (approved):

PR #424, opened 2025-08-07
feat: support markdown when renderAsPills is true

ctlai95:
## Problem

Custom renderer for pills doesn't support markdown

## Solution

- Add parseMarkdown helper to the customRenderer
- Also allow avoiding dispatching click events if `clickable` is explicitly set to false
image

## Tests

## License

### DIFF
--- a/src/components/chat-item/chat-item-card.ts
+++ b/src/components/chat-item/chat-item-card.ts
@@ -27,6 +27,7 @@ import { MoreContentIndicator } from '../more-content-indicator';
 import { Button } from '../button';
 import { Overlay, OverlayHorizontalDirection, OverlayVerticalDirection } from '../overlay';
 import { marked } from 'marked';
+import { parseMarkdown } from '../../helper/marked';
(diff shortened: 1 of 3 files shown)


src/components/chat-item/chat-item-card.ts
```diff
@@ -279,6 +316,9 @@ export class ChatItemCard {
         children: [ fileName ],
         events: {
           click: () => {
+            if (header.fileList?.details?.[filePath]?.clickable === false) {
```
laileni-aws, line 319: For my understanding, with this code change,
- User can click the pill and we return without any action or
- We don't show any experience of click on the pill?

ctlai95: This is to allow clicking on the pills to be **conditionally** disabled.

What that means is that for some cases where pills represent folders we can choose to disable the click events, while for other cases where pills represent files we can enable clicks (to open in the editor for example)

PR #489, opened 2026-07-15
fix: model selector tooltip renders raw HTML tags instead of text

laileni-aws:
## Summary

The model selector's hover tooltip displayed raw HTML tags (e.g. `<strong>` and `<br>`) as literal text instead of a formatted label and description. This makes the tooltip look broken to users. This PR fixes the tooltip to render correctly and adds unit tests to guard against regression.

## Problem

When hovering over a `select` form item that has options with a `description` (such as the model selector in chat), the tooltip showed something like:

```
<strong>Claude Sonnet 4</strong><br>Hybrid reasoning and coding for regular use
```

instead of a bold label followed by its description.

### Before

Screenshot 2026-09-15 at 12 30 20 PM
Screenshot 2026-09-15 at 12 30 13 PM

### After
Screenshot 2026-09-15 at 12 15 45 PM
Screenshot 2026-09-15 at 12 15 38 PM

## Root cause

`SelectInternal.getCurrentTooltip()` in `src/components/form-items/select.ts` built the tooltip content using raw HTML tags:

```ts
return `<strong>${selectedOption.label}</strong><br>${selectedOption.description}`;
```

That string is passed to `showTooltip()`, which renders it through a `CardBody`. `CardBody` renders its `body` via the markdown parser (`parseMarkdown`), and the parser's HTML renderer is configured to escape raw HTML:

```ts
// src/helper/marked.ts -> configureMarked()
html: ({ text }) => escapeHTML(text),
```

As a result, the raw `<strong>` / `<br>` tags were escaped and shown to the user verbatim.

## Fix

Use markdown syntax instead of raw HTML so the content renders correctly through the same parser:

```ts
return `**${selectedOption.label}**\n${selectedOption.description}`;
```

`CardBody` already invokes the parser with line breaks enabled (`includeLineBreaks: true`), so the newline becomes a line break and `**…**` renders as bold — producing the intended "bold label + description" tooltip without any raw HTML.

## Testing

- Added unit tests in `src/__test__/components/form-items/select.spec.ts` covering the tooltip content path:
  - the tooltip renders as markdown (a real `<strong>` element) and never leaks literal `<strong>` / `<br>` text;
  - the tooltip reflects the currently selected option;
  - it falls back to the base `tooltip` when the selected option has no description.
- The new tests were confirmed to fail against the previous (raw-HTML) implementation and pass with this change.
- Full unit test suite passes (`npx jest`): 72 suites / 845 tests.
- `eslint` and `prettier --check` pass; production build (`npm run build`) succeeds.

## Additional change

- `ui-tests/__test__/flows/quick-action-commands-header.ts`: replaced an implicit truthiness check on an `any`-typed value with an explicit boolean comparison to satisfy `@typescript-eslint/strict-boolean-expressions`. This was a pre-existing lint error unrelated to the tooltip fix; correcting it keeps the lint gate green.

## Notes

- Change is limited to tooltip content generation; no public API changes.

### DIFF
--- a/ui-tests/__test__/flows/quick-action-commands-header.ts
+++ b/ui-tests/__test__/flows/quick-action-commands-header.ts
@@ -140,7 +140,7 @@ export const verifyQuickActionCommandsHeaderStatusVariations = async (page: Page
     const hasStatus = await headerElement.evaluate((el, className) =>
       el.classList.contains(className), statusClass
     );
-    if (hasStatus) {
+    if (hasStatus === true) {
       foundStatusClass = true;
       console.log(`Found status class: ${statusClass}`);
       break;
--- a/ui-tests/__test__/flows/quick-action-commands-header.ts
+++ b/ui-tests/__test__/flows/quick-action-commands-header.ts
@@ -137,10 +137,10 @@ export const verifyQuickActionCommandsHeaderStatusVariations = async (page: Page
   let foundStatusClass = false;
 
   for (const statusClass of statusClasses) {
-    const hasStatus = await headerElement.evaluate((el, className) =>
+    const hasStatus = Boolean(await headerElement.evaluate((el, className) =>
       el.classList.contains(className), statusClass
-    );
-    if (hasStatus === true) {
+    ));
+    if (hasStatus) {
       foundStatusClass = true;
       console.log(`Found status class: ${statusClass}`);
       break;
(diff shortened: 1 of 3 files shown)


ui-tests/__test__/flows/quick-action-commands-header.ts
```diff
@@ -137,9 +137,9 @@ export const verifyQuickActionCommandsHeaderStatusVariations = async (page: Page
   let foundStatusClass = false;
 
   for (const statusClass of statusClasses) {
-    const hasStatus = await headerElement.evaluate((el, className) =>
+    const hasStatus = Boolean(await headerElement.evaluate((el, className) =>
```
dungdong-aws: this is unrelated to the ticket

dungdong-aws (approved):

chungjac (approved):

PR #492, opened 2026-07-15
fix: preserve undo history when pasting into the chat input

laileni-aws:
## Summary

Pasting text into the chat prompt input broke undo: pressing Ctrl/Cmd+Z after a paste did not undo the pasted text correctly. This PR inserts pasted text in a way that participates in the browser's native undo history.

## Problem

The prompt input's `paste` handler prevented the default paste and inserted the clipboard text by manually creating and inserting a DOM text node at the cursor:

```ts
e.preventDefault();
const text = e.clipboardData?.getData('text/plain');
// ...
range.insertNode(document.createTextNode(text));
```

Directly mutating the DOM this way does not add an entry to the browser's undo stack, so after a paste the editor's undo history is inconsistent and Ctrl/Cmd+Z does not undo the paste as expected.

## Fix

Insert the pasted text with `document.execCommand('insertText', false, text)`, which performs the insertion through the editing pipeline and records it in the native undo history. If `execCommand` is unavailable (or returns `false`), fall back to the previous manual text-node insertion so behavior is preserved in environments that do not support it.

```ts
let inserted = false;
try {
  inserted = document.execCommand('insertText', false, text);
} catch {
  inserted = false;
}
if (!inserted) {
  // fallback: manual insertion at the cursor
}
```

## Testing

- Added unit tests in `src/components/__test__/chat-item/prompt-text-input.spec.ts`:
  - paste inserts text via `execCommand('insertText')` (the undo-preserving path) and fires the input event;
  - when `execCommand` is unavailable, the handler falls back to manual insertion and the text still appears;
  - paste events with no text are ignored.
- The `execCommand` test was confirmed to fail against the previous manual-insertion implementation and pass with this change.
- Full unit test suite passes (`npx jest`): 844 tests.
- `eslint` and `prettier --check` pass; production build (`npm run build`) succeeds.

## Additional change

- `ui-tests/__test__/flows/quick-action-commands-header.ts`: coerced the result of a Playwright `evaluate` call to a boolean with `Boolean(...)` before using it in a condition, satisfying `@typescript-eslint/strict-boolean-expressions` and keeping the repo lint gate green.

## Notes

- `document.execCommand` is deprecated but remains the only widely supported way to insert text into a `contenteditable` element while preserving the native undo stack; the fallback keeps the previous behavior where it is not available.

### DIFF
--- a/src/components/__test__/chat-item/prompt-text-input.spec.ts
+++ b/src/components/__test__/chat-item/prompt-text-input.spec.ts
@@ -112,7 +112,44 @@ describe('prompt-text-input', () => {
     expect(inputElement?.getAttribute('placeholder')).toBe('Updated placeholder');
   });
 
-  it('handles paste events', () => {
+  it('inserts pasted text via execCommand so it can be undone', () => {
+    const testTabId = MynahUITabsStore.getInstance().addTab({
+      isSelected: true,
+      store: {}
+    }) as string;
+
+    let inputCalled = false;
+    const textInput = new PromptTextInput({
+      tabId: testTabId,
+      initMaxLength: 1000,
+      onKeydown: () => {},
+      onInput: () => { inputCalled = true; }
+    });
+
+    const inputElement = textInput.render.querySelector('.mynah-chat-prompt-input') as HTMLElement;
+
+    const originalExecCommand = document.execCommand;
+    const execCommandMock = jest.fn(() => true);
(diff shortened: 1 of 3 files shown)


dungdong-aws:
this fix only works in mynah-ui itself, won't work in a webview in vscode. Raised a new PR: https://github.com/Amazon-Q-Developer/mynah-ui/pull/498

PR #499, opened 2026-09-15
fix: render button tooltips through CardBody once

dungdong-aws:
## Problem

Icon-only buttons in the top navigation bar — including MCP and chat history — opened a tooltip bubble with no usable text.

### Before
Screenshot 2026-09-15 at 4 36 02 PM

Screenshot 2026-09-15 at 4 36 13 PM

### After
Screenshot 2026-09-15 at 4 36 39 PM

Screenshot 2026-09-15 at 4 36 46 PM

## Root cause

Button tooltips were parsed twice:

1. `Button` converted the label/tooltip markdown to HTML with `parseMarkdown(...)`.
2. `Button` passed that generated HTML to `CardBody`, which owns markdown rendering and parsed it again.

This double parse was latent from the original Button tooltip implementation (`1d733b7c`, March 2024). It remained harmless while marked passed raw HTML through, including when the MCP/history descriptions were present in July 2025 with mynah-ui 4.36.2.

The XSS hardening in #484 / `571d6c25` correctly began escaping all raw HTML. From mynah-ui 4.40.2 onward, the second parse therefore treated the first parse's generated `<p>...</p>` as untrusted HTML, exposing the latent defect. The security protection should remain unchanged.

## Fix

Make `CardBody` the single owner of markdown rendering. `Button` now assembles and passes raw label/tooltip markdown instead of pre-rendered HTML.

```diff
- tooltipText += parseMarkdown(props.tooltip ?? '', { includeLineBreaks: true });
+ tooltipText += props.tooltip;
```

This preserves markdown formatting while avoiding the security-hardened second parse.

## Tests

Added focused coverage for both affected paths:

- explicit tooltip markdown on an icon-only button
- automatically generated tooltip markdown for a truncated button label

Each test captures the content passed through the tooltip Overlay and verifies it was rendered exactly once, including a real `<strong>` element rather than escaped generated HTML.

## Validation

- Verified manually in the mynah-ui example app: the previously empty MCP/top-navigation tooltip now displays its text.
- PR CI runs lint, unit-tests and e2e-linux.
- Local automated tests were not run; CI is the first execution of the new tests.


src/components/button.ts
```diff
@@ -174,15 +174,16 @@ class ButtonInternal extends ButtonAbstract {
...
+            tooltipText = props.label;
           }
           if (props.tooltip !== undefined) {
             if (tooltipText != null) {
               tooltipText += '\n\n';
             } else {
               tooltipText = '';
             }
-            tooltipText += parseMarkdown(props.tooltip ?? '', { includeLineBreaks: true });
+            // CardBody owns markdown rendering; pre-rendering here causes its parser to escape the generated HTML.
```
laileni-aws, line 185: Instead we should not send tooltip option at all from the LS? keep this tooltip experience as optional

dungdong-aws: As optional, do you mean by if LS does not send it, we should not render anything at all instead of an empty bubble?

laileni-aws: Yes, LS should decide to show tooltip or not.

dungdong-aws: actually it did check for `null` description but it did not check for empty string. I added it so it won't render empty string as well

